PR #1, opened 2021-06-28
Updated dependencies with the new gtk release.

MGlolenstine:
I've updated dependencies, as the [gtk4](https://crates.io/crates/gtk4) has been released on crates.io.

~~I've also had to add `use gtk4 as gtk;` to avoid compilation problems where it was expecting `gtk` and found `gtk4`.
This could be solved by rewriting all dependencies to the `gtk4`, but this was a fast and hacky solution.~~
I've rewritten all of `gtk::`, as it was a matter of just replacing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,22 +1,18 @@
 [package]
-name = "relm4"
-version = "0.1.0"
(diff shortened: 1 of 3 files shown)


AaronErhardt (approved):
Thanks again!

PR #3, opened 2021-07-26
Updated gtk4 to 0.2 and new declaration.

MGlolenstine:
The GTK4 has been updated to 0.2, and [apparently](https://gtk-rs.org/gtk4-rs/git/book/prerequisites.html#cargo) a new way of specifying the package.

I haven't encountered any issues when issuing `cargo build` but haven't tested it on any projects.


AaronErhardt:
Thanks! The new dependency declaration looks much better IMO.

PR #14, opened 2021-08-21
Remove classname methods from WidgetPlus

overlisted:
There is already `WidgetExt::{add_css_class, remove_css_class}` in GTK4.


AaronErhardt (approved):
Thanks for the PR! I'm glad you noticed that.

PR #18, opened 2021-09-04
typo

tronta:


AaronErhardt:
Looks good. You need to use element instead of elements though :)

tronta:
thanks for the hint :smile: 
is fixed

AaronErhardt:
Perfect :+1:

PR #19, opened 2021-09-04
Would it make sense to simplify this?

tronta:


AaronErhardt:
You're right, of course, that this doesn't make sense. But the reason it is written like that is to give an example of this feature and to also test the macro. Relm4 has no tests in itself as of now, but is tested against the examples in the CI so if the macro would have a regression with optional features it'd be detected.

tronta:
Ah, I see, sorry, got it. You are right.

Issue #25, opened 2021-09-14
Issue with reusing components twice


mskorkowski:
I've created a fairly complex reusable component which I would like to use twice with different settings in the same view. The way in  which components are initiated makes it hard to select proper configuration of the component.

I've simplified the case and created a gist with example issue and solution I came up with. https://gist.github.com/mskorkowski/ae88013c4d19b6479270aeeaa08e8416

Current state of relm4 makes it hard to reuse components more then once in the widget.

I don't like the solution I've created because
1. It moves configuration of the component away from the creation step (`Components::init_components`)
2. Overly complex and verbose
3. Produces polution in the namespace if component is used many times in application

What I would like to, is 
1. `ComponentUpdate::init_model` to take component settings struct as an argument instead of parent model
2. `RelmComponent::new` had additional argument for component settings struct

AaronErhardt:
Thanks for your feedback!
Maybe your solution can be simplified. Generics should work pretty well in this case. I will try it out this afternoon. If that doesn't work we can still use your suggestion.

mskorkowski:
I've created a pull request with implementation where the settings are passed alongside the model. In this solution, Model has another dependant type.

```
struct Model {
    type Msg: 'static;
    type Widgets;
    type Components: Components<Self>;
    type Settings;
}
```

https://github.com/AaronErhardt/relm4/pull/26

Things I found nice about this approach is that
1. Every part of application can receive initial configuration in plain way
2. `RelmApp::new` also needs a Settings, which could allow in future setting proper application id for apps using relm as per gtk requirement.

AaronErhardt:
Thanks a lot for your effort!

I appreciate your idea, but to be honest, I'm not really convinced by your approach yet. The problem is that adding the `Settings` type to the `Model` trait is only useful for reusable components and nothing else.

Regarding point 1, the model for the app is already initialized from outside and the components have the `init_model` function for exactly this
Regarding point 2, I don't see how that would help with the application id. Also there's this already: https://aaronerhardt.github.io/docs/relm4/relm4/fn.gtk_application.html

I've tried to make the alert component reusable twice or more without changing the relm4 crate. I've added the changes to this branch: https://github.com/AaronErhardt/relm4/tree/reusable-comps
IMO this approach is pretty reasonable and it only required a few extra lines of code both for the component and the application part.

mskorkowski:
> Regarding point 1, the model for the app is already initialized from outside and the components have the init_model function for exactly this
[...]

PR #26, opened 2021-09-14
[WIP] Make reusing components in the same view is easier

mskorkowski:
Model contains new dependant type `Settings`. This
type is passed around so it's easy to differentiate
multiple compnents configuration, dependant on the call.


mskorkowski:
Proposal for #25

mskorkowski:
As discussed in ticket, we decided to not go this way

PR #29, opened 2021-09-19
examples: make tracker example more idiomatic

Hofer-Julian:
It now uses an iterator to select icons in random


AaronErhardt:
Thanks!

PR #31, opened 2021-09-23
add new entry example

tronta:
What do you think about this?


AaronErhardt:
I really like it! Clever use of the track! macro :+1:

Could you maybe use if-else instead of match for positive and negative numbers to make it run on Rust 1.54?

Maybe you can even use this in the if-else conditions: https://doc.rust-lang.org/std/primitive.i32.html#method.is_positive

tronta:
I noticed that I cleared unnecessarily `entry` at three placed in `update`

AaronErhardt:
Thanks!

PR #32, opened 2021-09-24
Reexport factory::FixedPosition

songww:


AaronErhardt:
Thanks for pointing this out. I already noticed this issue but I'm currently working on improving factories in general for the 0.2 release so that they work better with containers like `adw::TabView`.

I hope you don't mind me closing this PR but as I said I'm already working on a mini re-work of factories and this will be fixed in the process as well :)

PR #33, opened 2021-09-24
Update entry example

AaronErhardt:
@tronta that's what I meant with storing the entry buffer in the model. That's nice because you don't have to send the input through the message and you also don't have to update the widgets.

What do you think?

I think you've also found a bug in FactoryVec. Have you tried to add several counters at once? It's did weird things when I tried to add 2+ elements.


tronta:
Nice! I wouldn't guessed that.

tronta:
> I think you've also found a bug in FactoryVec. Have you tried to add several counters at once? It's did weird things when I tried to add 2+ elements.

Worked for me. It just crashed for bigger values due to memory limitations.

tronta:
If you want we can take your version. If you think it makes sense we could also use variant a and b to show the `track!` functionality.

AaronErhardt:
Both examples are now available, the PR is no longer necessary.

PR #35, opened 2021-10-02
add new example

tronta:
What do you think about this as an example?
Does it provide enough new things for the user?
Can it be improved?

Some tings I want to improve anyway. Eg how calculations get generated.


AaronErhardt:
Thanks a lot! I'll have a closer look at this soon. 

Btw. examples that rely on libadwaita live in `relm4-examples/libadwaita/examples` because I don't want to have libadwaita as additional dependency for people who want to try regular examples.

tronta:
True, I was not really aware of using adwaita :grinning:

tronta:
I update my OS and now I get that error
```
error[E0599]: the function or associated item `set_child` exists for trait object `(dyn libadwaita::traits::ApplicationWindowExt + 'static)`, but its trait bounds were not satisfied
   --> src/main.rs:108:35
    |
108 |             ApplicationWindowExt::set_child: main_box = Some(&gtk::Box) {
    |                                   ^^^^^^^^^ function or associated item cannot be called on `(dyn libadwaita::traits::ApplicationWindowExt + 'static)` due to unsatisfied trait bounds
    | 
```
Maybe you can give me a hint, what to change?

AaronErhardt:
The method was renamed to `set_content` to avoid naming conflicts: https://world.pages.gitlab.gnome.org/Rust/libadwaita-rs/stable/latest/docs/libadwaita/prelude/trait.ApplicationWindowExt.html#tymethod.set_content

This also means that the trait disambiguation (the `ApplicationWindowExt::` part) can be removed because Rust knows which method comes from which trait.

I've updated the other libadwaita example as well: https://github.com/AaronErhardt/relm4/commit/e30b678b575e33143c4169db8517d4de51feacf5

tronta:
Thanks a lot. I was mainly confused because I was still pointing to alpha-3 cargo.toml.
`adw = { version = "0.1.0-alpha-3", package = "libadwaita" }`

tronta:
If you want, you can have a look again.
What do you think?

AaronErhardt:
Looks really cool. I like it! It's really fun to play!

tronta:
Since I'm easily excited about this kind of thing, I bought a book on quick math tricks. Who knows, maybe I'll do a follow-up project :smile:

PR #37, opened 2021-10-04
Named visibility argument for widget macro

mskorkowski:
Adds support for named argument for widget visibility.

Widget macro gains it's first named argument. This is a step forward to allow overriding the relm4 and gtk in widget macro.

### DIFF
--- a/relm4-macros/src/attrs.rs
+++ b/relm4-macros/src/attrs.rs
@@ -1,13 +1,26 @@
+use proc_macro2::Delimiter;
 use proc_macro2::TokenStream as TokenStream2;
 use quote::ToTokens;
-use syn::{
-    parse::{Parse, ParseStream},
-    Result, Visibility,
-};
+use syn::Ident;
+use syn::Result;
+use syn::Token;
+use syn::Visibility;
+use syn::buffer::Cursor;
+use syn::parse::Parse;
+use syn::parse::ParseStream;
+use syn::token;
 
 #[derive(Debug)]
 pub struct WidgetVisibility {
     pub_vis: Option<Visibility>,
+}
+
+impl WidgetVisibility {
+    fn new() -> Self {
+        WidgetVisibility{
+            pub_vis: None,
+        }
+    }
 }
 
 impl ToTokens for WidgetVisibility {
@@ -20,21 +33,85 @@ impl ToTokens for WidgetVisibility {
 
 pub struct Attrs {
     pub visibility: WidgetVisibility,
+    visibility_set: bool,
+}
+
+impl Attrs {
+    fn new() -> Self {
+        Attrs {
+            visibility: WidgetVisibility::new(),
+            visibility_set: false,
+        }
+    }
+
+    fn stream_peek(c: Cursor) {
+        if !c.eof() {
+            //group it is not
+            if let Some( (ident, cursor) ) = c.ident() {
+                eprintln!("\t ident: {}", ident);
+                return Attrs::stream_peek(cursor);
+            }
+
+            if let Some( (punct, cursor) ) = c.punct() {
+                eprintln!("\t punct: {}", punct);
+                return Attrs::stream_peek(cursor);
+            }
+
+            if let Some( (literal, cursor) ) = c.literal() {
+                eprintln!("\t literal: {}", literal);
+                return Attrs::stream_peek(cursor);
+            }
+
+            if let Some( (lifetime, cursor) ) = c.lifetime() {
+                eprintln!("\t lifetime: {}", lifetime);
+                return Attrs::stream_peek(cursor);
+            }
+
+            if let Some( (tree, cursor) ) = c.token_tree() {
+                eprintln!("\t tree: {}", tree);
+                Attrs::stream_peek(cursor);
+            }
+        }
+    }
 }
 
 impl Parse for Attrs {
+    /// Rules for parsing attributes
+    /// 
+    /// 1. It's fine if visibility is used unnamed so `#[widget(pub)]` must be valid
+    /// 2. Widget visibility might be named `#[widget(visibility = pub)]`
+    ///
     fn parse(input: ParseStream) -> Result<Self> {
-        let pub_vis = if input.is_empty() {
-            None
-        } else {
-            Some(input.parse()?)
-        };
+        eprintln!("Input:");
+        eprintln!("\tis empty: {}", input.is_empty());
 
-        let visibility = WidgetVisibility{
-            pub_vis,
-        };
+        let mut attrs = Attrs::new();
 
-        Ok(Attrs { visibility })
+        Attrs::stream_peek(input.cursor());
+
+        if input.peek(Token![pub]) {
+            let pub_vis = if input.is_empty() {
+                None
+            } else {
+                Some(input.parse()?)
+            };
+            attrs.visibility.pub_vis = pub_vis;
+        }
+        else if input.peek(Ident) && input.peek2(Token![=]){
+            let ident: Ident = input.parse()?;
+            let eq: Token![=] = input.parse()?;
+
+            let pub_vis = if input.is_empty() {
+                None
+            } else {
+                Some(input.parse()?)
+            };
+            attrs.visibility.pub_vis = pub_vis;
+        }
+
+        Ok(attrs)
     }
+
+    
 }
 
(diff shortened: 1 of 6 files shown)


relm4-macros/src/attrs.rs
```diff
@@ -1,31 +1,125 @@
...
+use syn::Ident;
+use syn::Result;
+use syn::Token;
+use syn::Visibility;
+use syn::parse::Parse;
+use syn::parse::ParseStream;
 
 #[derive(Debug)]
-pub struct Attrs {
+pub struct WidgetVisibility {
```
AaronErhardt, line 11: I don't think this struct is necessary. `Option<Visibility>` already implements ToTokens.

AaronErhardt:
Thank! Looks pretty good. 
I'll have a closer look at this in a few hours.

relm4-macros/src/attrs.rs
mskorkowski: Removed

relm4-macros/src/attrs.rs
```diff
@@ -1,84 +1,111 @@
...
-        if let Some(vis_pub) = &self.pub_vis {
-            vis_pub.to_tokens(out);
-        }
-    }
-}
-
 pub struct Attrs {
     /// Keeps information about visibility of the widget
-    pub visibility: WidgetVisibility,
+    pub visibility: Option<Visibility>,
```
AaronErhardt, line 18: I think `visibility_set` can be removed and `visibility.is_some()` can be used instead.

relm4-macros/src/attrs.rs
```diff
@@ -93,28 +120,65 @@ impl Parse for Attrs {
...
-                None
+        while !input.is_empty() {
+
+            // Attrs::stream_peek(input.cursor());
+
+            if input.peek(Token![pub]) {
+                if attrs.visibility_set {
+                    return Err(input.error("You can't assign visibility twice"));
+                }
+                let pub_vis = if input.is_empty() {
```
AaronErhardt, line 131: This can never be `true`, right? Otherwise the `while` condition would be `false`. 

Also IMO we don't need backwards comparability. Using `visibility = pub` is just much clearer than `pub` especially since we can have multiple arguments now.

relm4-macros/src/attrs.rs
```diff
@@ -93,28 +120,65 @@ impl Parse for Attrs {
...
+                    Some(input.parse()?)
+                };
+                attrs.visibility = pub_vis;
+                attrs.visibility_set = true;
+            }
+            else if input.peek(Ident) && input.peek2(Token![=]){
+                // This allows to report errors at the argument name instead of value
+                let visibility_error = Err(input.error("You can't assign visibility twice"));
+                let relm4_error = Err(input.error("You can't assign relm4 path twice"));
+                let unknown_identifier_error = Err(input.error("Unknown argument. Valid arguments are: `visibility`, `relm4` and `gtk` "));
```
AaronErhardt, line 143: I would prefer creating the errors when they are actually needed.

relm4-macros/src/attrs.rs
```diff
@@ -93,28 +120,65 @@ impl Parse for Attrs {
...
+
+                let ident: Ident = input.parse()?;
+                let _eq: Token![=] = input.parse()?;
+
+                let ident_value: String = ident.to_string();
```
AaronErhardt, line 148: You can compare `Ident` with `&str` even without converting it to a sting before AFAIK.

relm4-macros/src/attrs.rs
```diff
@@ -93,28 +120,65 @@ impl Parse for Attrs {
...
+
+                if ident_value == "visibility" {
+                    if attrs.visibility_set {
+                        return visibility_error;
+                    }
+                    let pub_vis = if input.is_empty() {
```
AaronErhardt, line 154: I don't think this makes sense. What would `visibility =` mean?

AaronErhardt (changes requested):
It's looking better and better. Just a few things to improve :)

relm4-macros/src/attrs.rs
mskorkowski: Sure, it's idiotic.

relm4-macros/src/attrs.rs
mskorkowski: > This can never be true, right? Otherwise the while condition would be false.

if you write `#[widget(visibility = pub, visibility = crate)] it can cause second assignment. It's stupid but having error here is cheap before somebody start to rely on undefined behavior (probably by mistake).

> Also IMO we don't need backwards comparability. Using visibility = pub is just much clearer than pub especially since we can have multiple arguments now.

I think rewriting relm4 due to reexports will be rather niche use case. While important for bigger projects. That's why I would keep option to use `#[widget(pub)]`. It's clean and compact. Most people would drift toward `visibility = pub` if they need to use `relm4` argument.

I propose the middle ground. If you use more then one argument you must use named args.

mskorkowski:
All should be fixed. @AaronErhardt can you review?

relm4-macros/src/attrs.rs
```diff
@@ -1,31 +1,176 @@
...
+use syn::PathArguments;
+use syn::PathSegment;
+use syn::Result;
+use syn::Token;
+use syn::Visibility;
+use syn::parse::Parse;
+use syn::parse::ParseStream;
+use syn::punctuated::Punctuated;
+use syn::spanned::Spanned;
+use syn::token::Colon2;
```
AaronErhardt, line 14: Maybe this can be written more compact. Like `syn::{Error, Ident, ...}`

relm4-macros/src/attrs.rs
```diff
@@ -1,31 +1,176 @@
...
+                }
+                let pub_vis: Visibility = input.parse()?;
+
+                attrs_type = AttributeType::Unnamed{
+                    span: pub_vis.span(),
+                };
+                attrs.visibility = Some(pub_vis);
+                dangling_comma = false;
+            }
+            else if input.peek(Ident) && input.peek2(Token![=]){
```
AaronErhardt, line 110: I would rather just parse `Ident` and the `=` unconditionally. Then you don't need `dangling_comma` and the user gets an error like `expected identifier` automatically.

relm4-macros/src/attrs.rs
```diff
@@ -1,31 +1,176 @@
...
+impl Parse for Attrs {
+    /// Rules for parsing attributes
+    /// 
+    /// 1. It's fine if visibility is used unnamed so `#[widget(pub)]` must be valid but thats the only case
+    /// 2. Widget visibility might be named `#[widget(visibility = pub)]`
+    /// 3. `relm4` argument must be named. Always
+    ///
+    fn parse(input: ParseStream) -> Result<Self> {
+        let mut attrs = Attrs::new();
+        let mut attrs_type = AttributeType::None;
```
AaronErhardt, line 78: Maybe a bit over-engineered but seems ok :D

relm4-macros/src/attrs.rs
```diff
@@ -1,31 +1,176 @@
...
+                if ident == "visibility" {
+                    let pub_vis: Visibility = input.parse()?;
+
+                    if let AttributeType::Unnamed{ span } = attrs_type {
+                        return Err(Error::new(span, mixed_use_error_message));
+                    }
+                    if attrs.visibility.is_some() {
+                        //unwrap's here are safe since all spans are from the same file always
+                        let error_span = ident.span().join(eq.span).unwrap()
+                                            .join(pub_vis.span()).unwrap();
```
AaronErhardt, line 123: Does this even run on stable? Maybe the docs are wrong, though: https://docs.rs/proc-macro2/1.0.29/proc_macro2/struct.Span.html#method.join

AaronErhardt: In any case, you don't need all those spans. Just the pub_viz span should be enough to figure out the error.

AaronErhardt (changes requested):
:+1: Just a few more things

relm4-macros/src/attrs.rs
mskorkowski: I'm using this syntax to reduce the amount of conflicts. 

I'll make it compact if you prefer

AaronErhardt (approved):
Looks good. Is everything ready to merge?

mskorkowski:
From my side yes

PR #39, opened 2021-10-14
Don't we need to initialize adwaita?

tronta:
With that, at least the theme seems to be correct.


AaronErhardt:
Thanks for pointing this out! I was totally unaware of this method. Instead of fixing this locally I directly patched Relm4 to automatically use `adw::Application` (which does the same as `adw::init()`) when the `libadwaita` feature is enabled: https://github.com/AaronErhardt/relm4/commit/0c6b4f74c113b5e5cf0e8837c6a38287e949e8b9

Btw. it would be great IMO if you put the widgets in your example into a grid and made your example resizable. That'd be even more cool :)

PR #41, opened 2021-10-15
add multislelect for calculations mathods

tronta:
I added the possibility to select multiple calculation methods


AaronErhardt:
You can either add a warning label to show that no operation was selected or you override input of the use (always re-select an operation if it was the last active action).

AaronErhardt:
Looks good! I'm looking forward to the timer.

PR #42, opened 2021-10-16
Add timer to calc-trainer example

tronta:
Some first implementation for the timer.
With some help of a kind guy, I got an easy setting of the button working.

With the timer I'm kind of lost. I tried to add a message handler. Not sure if this makes sense. Maybe you could point me in the right direction.

### DIFF
--- a/relm4-examples/libadwaita/examples/calc-trainer.rs
+++ b/relm4-examples/libadwaita/examples/calc-trainer.rs
@@ -269,6 +269,8 @@ impl AppUpdate for AppModel {
                 if self.timer.is_some() {
                     self.timer = None;
                 } else {
+                    self.pick_random_task_type();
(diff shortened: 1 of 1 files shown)


AaronErhardt:
Using a message handler should be the right approach. The easiest solution would be to send a message to the parent every second an let the parent decide what to do (or to ignore it). You could also make the message handler more complex, e. g. by adding an AtomicBool that  allows you stop or continue the timer from outside. 
In either case I would just send a "one second passed" message without any additional info and then store the actual counter in the app.

tronta:
That's exactly what I tried, It's only not being triggered yet. Not sure why. Probably something for tomorrow or so.

tronta:
Sorry, but I give up adding a long living mutex_guard :cry: 
Can you give me a hint?

PR #47, opened 2021-11-01
New Adwaita View-Switcher Example

tronta:
I'm trying to create a new example for using a view switcher in adwaita.

### DIFF
--- a/relm4-examples/libadwaita/examples/view-switcher.rs
+++ b/relm4-examples/libadwaita/examples/view-switcher.rs
@@ -1,0 +1,79 @@
+use adw::traits::ApplicationWindowExt;
+use gtk::prelude::{BoxExt, ButtonExt, GtkWindowExt, OrientableExt};
+use relm4::{send, AppUpdate, Model, RelmApp, Sender, WidgetPlus, Widgets};
+
+#[derive(Default)]
+struct AppModel {
+    counter: u8,
+}
+
+enum AppMsg {
+    Increment,
+    Decrement,
+}
+
+impl Model for AppModel {
+    type Msg = AppMsg;
+    type Widgets = AppWidgets;
+    type Components = ();
+}
+
+impl AppUpdate for AppModel {
+    fn update(&mut self, msg: AppMsg, _components: &(), _sender: Sender<AppMsg>) -> bool {
+        match msg {
+            AppMsg::Increment => {
+                self.counter = self.counter.wrapping_add(1);
+            }
+            AppMsg::Decrement => {
+                self.counter = self.counter.wrapping_sub(1);
+            }
+        }
+        true
+    }
+}
+
+fn application_window() -> adw::ApplicationWindow {
+    adw::ApplicationWindow::builder().build()
+}
+
+#[relm4_macros::widget]
+impl Widgets<AppModel, ()> for AppWidgets {
+    view! {
+        main_window = application_window() -> adw::ApplicationWindow {
+            set_default_width: 300,
+            set_default_height: 200,
+
+            set_content = Some(&gtk::Box) {
+                set_orientation: gtk::Orientation::Vertical,
+                set_margin_all: 5,
+                set_spacing: 5,
+
+                append = &adw::HeaderBar {
+                    set_title_widget = Some(&adw::ViewSwitcherTitle) {
+                        set_title: "A stack switcher App",
+                        set_stack: Some(&stack),
+                    }
+                },
+                
+                append: stack = &adw::ViewStack {
+                        add_titled: args!(
+                            &gtk::Label {
+                            set_label: "Start",
+                        }, Some("Title"), "title"),
+                        add: intro_label = &gtk::Label {
+                            set_label: "To get started click on \"Next page\".",
+                        },
+                        add: overview_label = &gtk::Label {
+                            set_label: "* Insert overview here *",
+                        },
+                }
+            },
+        }
+    }
+}
+
+fn main() {
+    let model = AppModel::default();
+    let app = RelmApp::new(model);
+    app.run();
+}
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
++ relm4-macros: New shorthand component syntax
+
 ## 0.2.1 - 2021-10-17
 
 ### Added
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 ### Added
 
++ relm4: Added sender method to RelmComponent
 + relm4-macros: Allow generic function parameters in properties
 
 ### Changed
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,11 @@
 # Changelog
 
-## Unreleased
+## 0.2.1 - 2021-10-17
 
 ### Added
 
 + relm4: Added sender method to RelmComponent
++ relm4-macros: New shorthand tracker syntax
 + relm4-macros: Allow generic function parameters in properties
 
 ### Changed
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "relm4"
 authors = ["Aaron Erhardt <aaron.erhardt@t-online.de>"]
 edition = "2018"
-version = "0.2.0"
+version = "0.2.1"
 resolver = "2"
 description = "An idiomatic GUI library inspired by Elm and based on gtk4-rs"
 repository = "https://github.com/AaronErhardt/relm4"
--- a/relm4-components/Cargo.toml
+++ b/relm4-components/Cargo.toml
@@ -2,7 +2,7 @@
 name = "relm4-components"
 authors = ["Aaron Erhardt <aaron.erhardt@t-online.de>"]
 edition = "2018"
-version = "0.2.0"
+version = "0.2.1"
 description = "An idiomatic GUI library inspired by Elm and based on gtk4-rs"
 repository = "https://github.com/AaronErhardt/relm4"
 readme = "README.md"
--- a/relm4-examples/examples/components.rs
+++ b/relm4-examples/examples/components.rs
@@ -189,7 +189,7 @@ impl Widgets<AppModel, ()> for AppWidgets {
         main_window = gtk::ApplicationWindow {
             set_default_width: 500,
             set_default_height: 250,
-            set_titlebar: component!(Some(components.header.root_widget())),
+            set_titlebar: component!(Some(header)),
             set_child = Some(&gtk::Label) {
                 set_label: watch!(&format!("Placeholder for {:?}", model.mode)),
             },
--- a/relm4-examples/examples/macro_test.rs
+++ b/relm4-examples/examples/macro_test.rs
@@ -112,7 +112,7 @@ impl Widgets<AppModel, ()> for AppWidgets {
                 set_margin_all?: Some(5),
                 set_spacing: 5,
 
-                append: component!(components.button1.root_widget()),
+                append: component!(button1),
                 append: inc_button = &gtk::Button {
                     set_label: "Increment",
                     connect_clicked(sender) => move |_| {
--- a/relm4-examples/examples/macro_test.rs
+++ b/relm4-examples/examples/macro_test.rs
@@ -4,6 +4,7 @@ use relm4::{
     WidgetPlus, Widgets,
 };
 
+#[tracker::track]
 struct AppModel {
     counter: u8,
     classes: Vec<&'static str>,
@@ -28,13 +29,14 @@ impl AppUpdate for AppModel {
         _components: &AppComponents,
         _sender: Sender<AppMsg>,
     ) -> bool {
+        self.reset();
         match msg {
             AppMsg::Increment => {
-                self.counter = self.counter.wrapping_add(1);
+                self.set_counter(self.counter.wrapping_add(1));
                 self.decrement = false;
             }
             AppMsg::Decrement => {
-                self.counter = self.counter.wrapping_sub(1);
+                self.set_counter(self.counter.wrapping_sub(1));
                 self.decrement = true;
             }
         }
@@ -135,7 +137,7 @@ impl Widgets<AppModel, ()> for AppWidgets {
                     set_column_spacing: 10,
                     set_column_homogeneous: true,
                     attach(1, 1, 1, 1) = &gtk::Label {
-                        set_label: "grid test 1",
+                        set_label: track! {&model.counter.to_string()},
                     },
                     attach(1, 2, 1, 1) = &gtk::Label {
                         set_label: "grid test 2",
@@ -175,6 +177,7 @@ fn main() {
         counter: 0,
         classes: vec!["first", "second"],
         decrement: false,
+        tracker: 0,
     };
     let app = RelmApp::new(model);
     app.run();
--- a/relm4-examples/examples/open_button.rs
+++ b/relm4-examples/examples/open_button.rs
@@ -98,7 +98,7 @@ impl Widgets<AppModel, ()> for AppWidgets {
             set_default_width: 300,
             set_default_height: 100,
             set_titlebar = Some(&gtk::HeaderBar) {
-                pack_start: component!(components.open_button.root_widget()),
+                pack_start: component!(open_button),
             }
         }
     }
--- a/relm4-examples/examples/stack_switcher.rs
+++ b/relm4-examples/examples/stack_switcher.rs
@@ -116,7 +116,7 @@ impl Widgets<AppModel, ()> for AppWidgets {
         main_window = gtk::ApplicationWindow {
             set_default_width: 500,
             set_default_height: 250,
-            set_titlebar: component!(Some(components.header.root_widget())),
+            set_titlebar: component!(Some(header)),
             set_child: component!(Some(components.stack.root_widget())),
         }
     }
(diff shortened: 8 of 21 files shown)


relm4-examples/libadwaita/examples/view-switcher.rs
```diff
@@ -0,0 +1,75 @@
...
+                append = &adw::HeaderBar {
+                    set_title_widget = Some(&adw::ViewSwitcherTitle) {
+                        set_title: "A stack switcher App",
+                        set_stack: Some(&stack),
+                    }
+                },
+                
+                append: stack = &adw::ViewStack {
+                    add_titled: args!(
+                        &gtk::Label::new(Some("This is the start page")), Some("First"), "First Page"),
```
tronta, line 60: Maybe you could help me on that:
I want to create those subpages, which use multiple arguments. Now I want to continue below with the macro notation. Do need to use components or such? Or is there something I am missing so far?

AaronErhardt: @tronta you should be able to use `add_titled(arg1, arg2, ...) = &gtk::Label ... `

tronta: The first parameter of the function is expecting the label, so I'm not sure how I can assign it as you proposed.

tronta: Indeed, it's working like this. You have probably described it somewhere, but I was not expecting this behavior. Maybe I forgot this. Thanks

relm4-examples/libadwaita/examples/view-switcher.rs
```diff
@@ -0,0 +1,78 @@
...
+                append = &adw::HeaderBar {
+                    set_title_widget = Some(&adw::ViewSwitcherTitle) {
+                        set_title: "A stack switcher App",
+                        set_stack: Some(&stack),
+                    }
+                },
+                
+                append: stack = &adw::ViewStack {
+                    add_titled(Some("First"), "First Page") = &gtk::Label {
+                        set_label: "This is the start page",
```
tronta, line 60: Sorry, I'm stuck. What would be the way to add icons and such?
To me the way how to untilize [set_icon_name](https://world.pages.gitlab.gnome.org/Rust/libadwaita-rs/stable/latest/docs/libadwaita/struct.ViewStackPage.html#method.set_icon_name) is still mysterious

AaronErhardt: Before using set_icon_name you either have to know how your theme calls a certain icon (which is not very reliable) or you add your own image as resource (there's an "Icon library" app to help you btw.). Using your own resources is a bit painful though because it usually requires meson as build system and a resource schema to load the image. It think you can also use resources with cargo and pure Rust but I haven't tried it yet.

An example code to load a resource files can be found here: https://gitlab.elektrotechnik.hs-augsburg.de/erhardt/state-factory/-/blob/gtk4/src/static_resources.rs (lines 22-24)

And together with this function you might be able to get images working without meson: https://gtk-rs.org/gtk-rs-core/git/docs/gio/fn.compile_resources.html

My idea would be: compile your XML schema (schemas look like [this](https://gitlab.elektrotechnik.hs-augsburg.de/erhardt/state-factory/-/blob/gtk4/data/resources/resources.gresource.xml)) -> load compiled schema -> register schema for your application -> use icon

tronta: Probably I should have been more precise with my problem description. I understand the concept with the named icons. The problem I have here is, how to create a `libadwaita::ViewStackPage` as it does not seam to have a constructor and somehow just gets implicitly created by using e.g. `add_titled`. But how can I later add  those icons to it?

tronta: An other question I have: Can I also combine `property_name(value1, value2, ...) = gtk::Box { ... }` with providing a name `property_name: name = gtk::Box { ... }`?
Maybe something like `property_name: name(value1, value2, ...) = gtk::Box { ... }`

AaronErhardt: The problem is that Relm4 doesn't handle cases yet where a property function returns a new widget. Libadwaita differs from GTK in this point and the macro will probably need a new syntax for this.

An yes you should be able to use a name for the widget. I think the syntax is like this:
`property(arg1, arg2, ...): name = gtk::Box { ... }`

tronta: Would it make sense to make them aware of this problem?
Maybe they could just add additional methods to avoid that problem. Or do you think it's not worth the effort?

AaronErhardt: I don't think they would introduce a new API for us unless it males sense in general. So if you have an idea for a new ViewStack API an issue would make sense. Otherwise Relm4 needs to adapt, I guess.

tronta: I would think about something like this
``` rust
                append: stack = &adw::ViewStack {
                    add: &adw::ViewStackPage {
                        set_child: = &gtk::Label {
                            set_label: "This is the start page",
                        },
                        set_icon_name: Some("icon_name"),
                        set_badge_number: 3,
                    },
                    add_titled(Some("First"), "First Page") = &gtk::Label {
                        set_label: "This is the start page",
                    },
```
What do you think about that? This would probably need two additional methods:

- `adw::ViewStackPage::new()`
- `adw::ViewStackPage::set_child(&self, child: &P)`

AaronErhardt: That's what I'd suggest as well. We need some sort of constructor and a method to set the child. Or even both at once. Do you have time to open an issue over at libadwaita?

tronta: Yes, currently I have time, but not sure, how the C-API looks like :smile: 
I'll will have a look.

AaronErhardt: You're basically looking for a new `view_stack_page_new` and `view_stack_page_set_child`method when "talking C".
More here: https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.0.0.alpha.4/

tronta: What do you think? https://gitlab.gnome.org/GNOME/libadwaita/-/issues/311

AaronErhardt: Looks good. Let's see what they think :)

AaronErhardt: Thanks for opening the issue! I already expected they wouldn't be happy about these changes (and to be honest I wouldn't be either in their place). That then means more work for me and a new syntax for handling these types of functions... suggestions are welcome :)

tronta:
What do you think about something like this?
``` Rust
                append: stack = &adw::ViewStack {
                    add_titled(Some("First"), "First Page") = &gtk::Label {
                        set_label: "This is the start page",
                    },
                    add_titled(Some("Second"), "Second Page") = &gtk::Label {
                        set_label: "This is the second page"
                    }.unwrap().set_badge_number(3),
                    v1 = add_titled(Some("Third"), "Third Page") = &gtk::Label {
                        set_label: "This is the last page"
                    }.unwrap().set_badge_number(3)
                }
```

AaronErhardt:
Mhh, maybe something like this? Using `returns` as a keyword to use the returned value and `?` to unwrap it.
```rust
                append: stack = &adw::ViewStack {
                    add_titled(Some("First"), "First Page") = &gtk::Label {
                        set_label: "This is the start page",
                    },
                    add_titled(Some("Second"), "Second Page") = &gtk::Label {
                        returns? = page: ViewStackPage {
                            set_badge_number: 3,
                        },
                        set_label: "This is the second page"
                    },
                }
```

tronta:
tbh I wouldn't expect the return value probably not inside the child definition, but its feasibility is of coarse important :smile: 

What about this?
```rust
                append: stack = &adw::ViewStack {
                    add_titled(Some("First"), "First Page") = &gtk::Label {
                        set_label: "This is the start page",
                    } -> page: ViewStackPage? {
                        set_badge_number: 3,
                    },
                    add_titled(Some("Second"), "Second Page") = &gtk::Label {
                        set_label: "This is the second page"
                    }
                }
```

tronta:
Great. How much effort is something like this? I have no clue. Haven't touched macro programming yet :disappointed:

AaronErhardt:
I don't know honestly. It's sometimes hard to tell when it comes to macros. In any case, it will probably take until the weekend before I'll be able to send a patch for this.

tronta:
Don't worry. Thanks a lot for you effort!

AaronErhardt:
Btw. I had a look at the macro. There might be a hack to implement this feature faster but I think it's time to re-organize the macro logic into a more understandable and extensible structure while adding this feature. 
That makes even more sense with micro components being added that require a new widget macro and I don't want to copy paste a lot of code.

tronta:
A solid code base sounds like a great idea!

AaronErhardt:
As an update on this matter: I've re-organized the macro structure to be much more clear now and able of handling the new syntax (actually parsing already works, but code generation is still missing). 

The problem is however, that the structure of Relm4 currently makes it impossible to handle this conveniently. Currently, both widgets and models are initialized from the root (the app) to the leaves (the outermost components). This is problematic because to keep the order of widgets and components in sync with the order they appear in the macro, the all widgets must be connected when the components are ready, which is after the widgets are actually created. Therefore the [connect_components method](https://aaronerhardt.github.io/docs/relm4/relm4/trait.Widgets.html#method.connect_components) exists to connect the components to the parents widgets in an extra step and is used by the macro to ensure correct ordering (see https://github.com/AaronErhardt/relm4/issues/20 for more context).

This was kind of a hack so far but worked fine until now. Because now we want to store a widget that is created by assigning a widget to another widget which happens in `connect_components`. But at this point the struct is already initialized which means that the new widget can't be part of the widgets struct any more which makes it hard to change the title of a ViewStackPage later on for example. 

TL;DR The current design of Relm4 would largely restrict the usability of the syntax we came up with.

As for a solution I've though about changing the order the widgets are created. The model should still be created from the root to the leaves, but I think maybe the widget can be created from the leaves to the root. With this, a parent component would already have access to its child component when creating its widgets so initialization and connecting the widgets can take place in one function which would solve the problem.

P. S. You don't need to bother about the technical details I wrote here, but I wrote this in case you @tronta or someone else is interested in this matter :)

tronta:
Hi @AaronErhardt, I'm really thankful that you keep me updated at this level of detail. Hope it does not impact too much the previous concept of Relm4.

AaronErhardt:
@tronta I've mostly finished my work on the rework-core branch. The new macro syntax already works now and only required a few changes to the core architecture. I just need to port relm4-components and several examples to the new version.

tronta:
What is the right way to apply a command on an optional result?
``` rust  
                  add_titled(Some("First"), "First Page") = &gtk::Label {
                        set_label: "This is the start page",
                    } ->  {
                        set_icon_name?: Some("")
                    },
```
something like this?
``` rust  
                  add_titled(Some("First"), "First Page") = &gtk::Label {
                        set_label: "This is the start page",
                    }? ->  {
                        set_icon_name?: Some("")
                    },
```

AaronErhardt:
@tronta I totally forgot that libadwaita returns `Option` there. I thought that optional values in this place wouldn't make sense because, well, how could this fail, right? So there is no syntax for this yet.

Even though it wouldn't be terribly complex to add such a syntax, I would rather suggest asking the maintainers of libadwaita-rs to make this non-optional because as far as I understand the C code this can not fail and shouldn't be optional.

tronta:
I guess you are right. I can create a ticket and ask. I'm trying to think about cases when this makes sense…

tronta:
https://gitlab.gnome.org/World/Rust/libadwaita-rs/-/issues/11

AaronErhardt:
Mhh, seems like this won't be fixed upstream soon, but still I don't quite understand their reasoning. They seems to have discovered the same thing over 4 months ago and still this API persists...

tronta:
Yes, indeed.

AaronErhardt:
@tronta I've added the syntax to the macro now. It should look like `-> ? { ... }`. I'll shortly release 0.4.0-beta.2 for this.

AaronErhardt:
0.4.0-beta.2 is out, let test it then :)

tronta:
What do you mean? Have you already implemented it? You are fast. What is the exact syntax?

AaronErhardt:
As I said, this wasn't really hard to add. I actually had the parsing part already but never implemented the code generation. 

So this is the syntax (question mark comes after the arrow):

```rust
                  add_titled(Some("First"), "First Page") = &gtk::Label {
                        set_label: "This is the start page",
                    } -> ? {
                        set_icon_name?: Some("")
                    },
```

Also, different from the normal optional assign this actually calls `unwrap()` and would panic. But that should be fine.

tronta:
Currently it seams like one needs to provide an explicit type of the returned value eg
``` rust
                    add_titled(Some("First"), "First Page") = &gtk::Label {
                        set_label: "This is the start page",
                    } -> first_page:ViewStackPage? {
                        set_icon_name: Some("eye-open-negative-filled-symbolic"),
                        set_needs_attention: true
                    },
```
would it be also possible to not provide the type like this?
``` rust
                    add_titled(Some("First"), "First Page") = &gtk::Label {
                        set_label: "This is the start page",
                    } -> first_page? {
                        set_icon_name: Some("eye-open-negative-filled-symbolic"),
                        set_needs_attention: true
                    },
```
As the return type is usually pretty defined.

AaronErhardt:
Yes, this is because the macro puts named returned widgets into the widgets struct. And there the code generation needs to generate a regular struct definition that requires a type. If it wasn't for the struct, Rust is smart enough to guess the type and therefore unnamed returned widgets that are not part of the struct don't need a type annotation.

```rust
Widgets {
    ...
    first_page: ViewStackPage,
}
```

tronta:
Just to understand it. The macro expands to this:
``` rust
        let first_page: ViewStackPage = stack
            .add_titled(&_gtk_label_2, Some("First"), "First Page")
            .unwrap();
        stack.add_titled(&_gtk_label_3, Some("Second"), "Second Page");
        stack.add_titled(&_gtk_label_4, Some("Third"), "Third Page");
[…]
        _adw_viewswitchertitle_0.set_title("A stack switcher App");
        _adw_viewswitchertitle_0.set_stack(Some(&stack));
        _gtk_label_2.set_label("This is the start page");
```
It would work equally if the type would not be provided
``` rust
        let first_page = stack
            .add_titled(&_gtk_label_2, Some("First"), "First Page")
            .unwrap();
        stack.add_titled(&_gtk_label_3, Some("Second"), "Second Page");
        stack.add_titled(&_gtk_label_4, Some("Third"), "Third Page");
[…]
        _adw_viewswitchertitle_0.set_title("A stack switcher App");
        _adw_viewswitchertitle_0.set_stack(Some(&stack));
        _gtk_label_2.set_label("This is the start page");
```
So I'm not sure, what you exactly  meant.

AaronErhardt:
Yes that's one part. But the type is actually only for the struct because the widget macro also defines the struct.

```rust
stuct AppWidgets {
    [...]
    first_page: ViewStackPage,
}

impl Widgets for AppWidgets {
[...]
    fn init_view() {
        let first_page: ViewStackPage = stack
            .add_titled(&_gtk_label_2, Some("First"), "First Page")
            .unwrap();
[...]
        first_page.set_icon_name(Some("eye-open-negative-filled-symbolic"));
        first_page.set_needs_attention(true);
[...]
        AppWidgets {
            [...]
            first_page,
        }
    }
}
```

tronta:
ah, thanks a lot. I'm still too tired too think properly. I missed that part.
Cool work.

tronta:
Still room for improvements, but maybe already good enough for merging,
What do you think?

tronta:
I'd like still to add, that the `ViewSwitcherTitle` switches automatically to the bottom when the width is too narrow.
I'd assume this works via
``` xml
          <object class="AdwViewSwitcherBar" id="switcher_bar">
            <property name="stack">stack</property>
            <property name="reveal" bind-source="switcher_title" bind-property="title-visible" bind-flags="sync-create"/>
          </object>
```

AaronErhardt:
The XML you copied here just sets the stack of the ViewSwitcherBar and binds the "reveal" property to the "title-visible" property of the title. This does only half of the job to make this work. What's done in the libadwaita demo is the following:

The title has two ViewSwitcherBars and one WindowTitle. The first, wide ViewSwitcherBar shows text next to the icons and the second, smaller one just the icons with a small text blow. And the WindowTitle just shows the title.

Then there is another (the third!) ViewSwitcherBar at the bottom. An that's where the property binding comes into play. All title widgets are placed in a Squeezer, that will show the largest of the three widgets that still fits on the screen. This causes the two ViewSwitcherBars to be hidden once the window is too small for them so the title is shown instead. And then the binding takes care of the ViewSwitcherBar at the bottom and shows is when the title is visible.

I hope that makes sense, I recommend running the adwaita demo with inspector to see this mechanism in action :)

AaronErhardt:
Thanks a lot for your work! I think, too, that this can be merged now :)
Sorry for the delay...

AaronErhardt:
@tronta here's the code for adding the bottom bar: https://github.com/AaronErhardt/relm4/commit/11cc880ef47b3f93cd8a9b623b08eb8de545b668

Issue #49, opened 2021-11-05
micro components


mskorkowski:
Let's take micro components discussion here.

To make live easier I'm pasting original comments about the micro component way of working here:

> I already mentioned "micro components". In this case two micro components would have to share state (using Rc<RefCell<Data>> which usually works fine). If micro components then had an update_view() method, you wouldn't even need to send a message but could just call that method.
For this a tracker that supports multiple views would also be interesting for efficient UI updates.

> Anyway, to make my idea more tangible, I've written down some Rust-like code that represents the concept of "micro-components". Micro-components are "loose" so they need the parent to attach the root widget to its widgets. They don't need to store a clone of their parents sender but have a "data" field that holds (currently immutable) data that can be used to store senders and other stuff according to the needs of the user (to make them more flexible). They can easily be modified by their parent component and also handle messages. The view function can be called manually by using update_view(). Of course there are some possible panics but they should be easy to debug and the only other option to RefCell would be unsafe code.
>
>What do you think? Would that make things easier for you? Is there a use-case that isn't properly covered?
>
> ```rust
> struct MicroComponent<Model: MicroModel> {
>    model: Rc<RefCell<Model>>,
>    widgets: Rc<RefCell<Model::Widgets>>,
>    data: Mode::Data, // for additional data such as senders to other components
>    sender: Sender<Model::Msg>,
> }
> 
> impl<Model: MicroModel> MicroComponent<Model> {
>    fn new(model: Model, data: Model::Data) -> Self { ... }
>    fn update_view(&self) { ... }
>    fn model(&self) -> &Model { ... }
>    fn model_mut(&self) -> &mut Model { ... }
>    fn widgets(&self) -> &Model::Widgets { ... }
>    fn send(&self, msg: Model::Msg) { ... }
>    fn sender(&self) -> Sender<Model::Msg> { ... }
> }
>
> trait MicroModel {
>    type Msg;
>    type Widgets: MicroWidgets<Self>;
>    type Data;
>    
>    fn update(&mut self, msg: Self::Msg, data: &Data, sender: Sender<Self::Msg>);
> }
>
> trait MicroWidgets<Model: MicroModel> {
>    type Root;
>    
>    fn init_view(&mut self, model: &Model, sender: Sender<Mode::Msg>) -> Self;
>    fn view(&mut self, model: &Model, sender: Sender<Mode::Msg>);
>    fn root_widget(&self) -> Self::Root;
> }
> ```

[...]

Issue #46, opened 2021-10-29
Data store for relm4


mskorkowski:
## What?

`Data store` is collection of business model records. To show the content of the `data store` you create a `data store view`. `data store view` just makes sure proper data from a store are visible. `data store` is responsible for owning records. Whenever `data store` is changed it notifies the view.

As I see it
- `Data store` is closest to worker
- `Data store view` is closest to factory

## Related issues

- #53 
- ~~#49~~

## Why?

1. I'm writing an application where I need to show tons of data (data set which is 100-200GB is norm)
2. I need synchronize view between components showing related state

My first attempt was to write custom factories, list views, components, workers, etc... It works fine except state management is super painful. I was passing tons of events between components which are far away to synchronize what is visible and how.  If I would like to test wherever all the messages are passed across components when I modify component tree is disaster.

Doing fairly incomplete implementation of the data store by extracting bits of my app allowed me to reduce amount of messages in system by around 60. Now in all places where I had managed to change the code to use `data store`/`data store view` I have only one dummy message for each component instance to trigger the redraw.

What's more factories provided by relm4 doesn't support grouping nor accessing the widgets in the factory which for example I need for toggle buttons.

## What's the issue?

I have to issues with my current code

1. This extra event per component routing required to trigger redraw
2. I've an issue with cyclic dependency required by components depending on the store view of the parent

Implementation of the `data store` and `data store view` is in super early phase but if you can give me some pointers how to make it cooperate better with relm4 I would be super grateful. 

The code for current version can be found in https://github.com/mskorkowski/relm4-store

### Issue 1: extra event per component routing required to trigger redraw

All examples are kind of simple todo application. 

If you run `cargo run --package=relm4-store-examples --example todo_3-early` it will simple to do application where list of todos is duplicated. If you modify the left list, right will be automatically updated and vice versa.

In the source code in `relm4-store-examples/examples/todo_3-early/view/main_window.rs` in implementation of `AppUpdate` there is 

```rust
impl AppUpdate for MainWindowViewModel {
    fn update(
        &mut self, 
        msg: Self::Msg , 
        components: &Self::Components, 
        _sender: Sender<Self::Msg>
    ) -> bool {
        match msg {
            MainWindowMsg::LeftListUpdate => {
                //force redraw of the components
                components.right_list.send(TaskMsg::Reload).unwrap();
            },
            MainWindowMsg::RightListUpdate => {
                //force redraw of the components
[...]

PR #50, opened 2021-11-05
Micro components

mskorkowski:
Implementation of #49 

It's required to make #46 go forward

### DIFF
--- a/src/micro_component.rs
+++ b/src/micro_component.rs
@@ -9,36 +9,51 @@ use std::fmt::Debug;
 use std::fmt::Display;
 use std::rc::Rc;
 
+/// Errors which might get returned from [`MicroComponent::update_view`] method
+/// 
+/// Both [`std::fmt::Display`] and [`std::fmt::Debug`] are implemented for this
+/// enum so if you just would like to show an error you don't need to do the
+/// matching.
 pub enum MicroComponentError {
+    /// Error returned if borrow failed
     Borrow(BorrowError),
+    /// Error returned if borrowing mutably failed
     BorrowMut(BorrowMutError),
 }
 
+/// Helper to convert values of [`std::cell::BorrowError`] into [`MicroComponentError`]
 impl From<BorrowError> for MicroComponentError {
     fn from(err: BorrowError) -> Self {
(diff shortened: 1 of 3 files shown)


src/micro_component.rs
```diff
@@ -0,0 +1,126 @@
+use gtk::glib::{self, Sender};
+
+use std::cell::Ref;
+use std::cell::RefCell;
+use std::cell::RefMut;
+use std::rc::Rc;
+
+pub struct MicroComponent<Model: MicroModel> {
+    model: Option<Rc<RefCell<Model>>>,
+    widgets: Option<Rc<RefCell<Model::Widgets>>>,
```
AaronErhardt, line 10: Looks good overall! I don't think those can be optional though.

src/micro_component.rs
```diff
@@ -0,0 +1,126 @@
...
+            model: Some(shared_model),
+            widgets: Some(shared_widgets),
+            root_widget,
+            data: shared_data,
+            sender: cloned_sender,
+        }
+    }
+    pub fn update_view(&self) { 
+        todo!()
+    }
```
AaronErhardt, line 74: The same as in the event loop. Try to acquire the widgets and the model and then call view. This function could theoretically return a `Result` instead of error handling through logs.

src/micro_component.rs
mskorkowski: Fixed

src/micro_component.rs
```diff
@@ -0,0 +1,234 @@
...
+                f.write_fmt(format_args!("{}", err))
+            }
+        }
+    }
+}
+
+/// Formats [`MicroComponentError`] for debug format `{:?}`
+/// 
+/// This allows you to print errors without doing `matching` or `if let` statements
+impl Debug for MicroComponentError {
```
AaronErhardt, line 54: Is there a specific reason to manually implement this? I think using `#derive[Debug]` should do the same.

AaronErhardt (approved):
Looks good and should be ready to merge. I will write an example for micro components and refine the API if necessary soon.

PR #55, opened 2021-11-28
Example divide

tronta:
I'm playing with an other example.


AaronErhardt:
Is it ready to merge? If not you can mark it as wip.

PR #60, opened 2021-12-19
do not pass CLI arguments to GTK by default

### LINKED ISSUE #59: Do not use gtk::Application's command-line argument handling by default
relm4 invokes `gtk::Application::run` which will attempt to automatically handle command-line arguments from the environment. This leads to the surprising behavior that if you supply *any* arguments to a relm4 application, you'll get a `Glib-GIO-CRITICAL` warning "This application can not open files." and the application will immediately exit.

I'd like to propose that `RelmApp::run` instead invokes `gtk::Application::run_with_args(&[])`, and provide a `RelmApp::run_with_args()` for users that want to opt-in to `gtk::Application`'s argument handling.

If this sounds reasonable, I'm happy to open a PR.

euclio:
Fixes #59.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -33,13 +33,23 @@ where
     Model::Components: Components<Model> + 'static,
 {
     /// Runs the application, returns once the application is closed.
+    ///
+    /// Unlike [`gtk::Application::run`](../gtk4/struct.Application.html#method.run), this function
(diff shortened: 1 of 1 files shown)


AaronErhardt (approved):
Looks good!

PR #68, opened 2022-01-05
fix doctest and run tests in CI

euclio:
While working on another change I noticed that this test was failing. It wasn't caught because it doesn't run in CI.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -42,12 +42,12 @@ jobs:
         run: cargo clippy -- --deny warnings
 
       # nightly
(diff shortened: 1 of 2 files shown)


AaronErhardt:
Thanks for the PR. Having tests in the CI is definitely a good idea but requires some extra code to get working as it run's on a server where GTK can't be initialized out of the box. But I think there are some good references like gtk-rs that got it working.

euclio:
Running under `xvfb` seems to be enough :)

AaronErhardt:
Thanks!

PR #75, opened 2022-01-15
add verses to calendar example

tronta:
I'm not an religious person, but I didn't want the example to be unfinished. :smile: 
Hope you like my verse selection.


AaronErhardt:
Well done, I couldn't have found better verses.
Also, I like how you and I did exactly one half of the verses :)

PR #77, opened 2022-01-16
Improve example

tronta:
Don't ask to solve x * 0 = 0 as it is hard to calculate.


AaronErhardt:
You're right, that's too much of a brain teaser :wink:

PR #83, opened 2022-01-23
Add sudoku example

tronta:
My initial version of my Sudoku solver.
Any ideas to make it better?
Maybe more Sudoku like experience…


relm4-examples/libadwaita/examples/sudoku_solver.rs
```diff
@@ -0,0 +1,197 @@
...
+                println!("Start solving with {content}");
+                let sudoku = Sudoku::from_str_line(content.as_str()).unwrap();
+
+                if let Some(solution) = sudoku.solve_one() {
+                    self.from_string(solution.to_string());
+                }
+            }
+            AppMsg::Clear => {
+                // Why does this not work
+                // self.values.clear();
```
tronta, line 78: Do you know, why this is not working? What would be the best way to set the initial values?

AaronErhardt: If `self.values.clear()` doesn't clear the list and also destroys all widgets that'd be a bug. Can you open an issue if that's the case? Nothing complicated, just as reminder.

relm4-examples/libadwaita/Cargo.toml
```diff
@@ -8,3 +8,5 @@ edition = "2021"
 rand = "0.8.4"
 relm4 = { path = "../../", features = ["libadwaita", "macros"] }
 relm4-components = { path = "../../relm4-components" }
+# For Sudoku example
+sudoku = "0.7.0"
```
tronta, line 12: Maybe there is a better way than polluting the common toml. 
Maybe it's ok. what do you think?

AaronErhardt: I think that's fine because sudoku seems to be a rather small dependency. But you can always add examples in their own crate (add a new folder and then `cargo init`) which also makes sense if you need multiple files.

relm4-examples/libadwaita/examples/sudoku_solver.rs
tronta: But after your suggestions, I think, the workaround is maybe not too bad, What do you think.

tronta:
The library solves btw also the "The world’s most difficult sudoku puzzle" easily :smile:

relm4-examples/libadwaita/examples/sudoku_solver.rs
AaronErhardt: Actually, if I see this currently, the workaround isn't a workaround after all. But there's a kinda big rework of Relm4 on the way and I might change factories to make them work with such code.

AaronErhardt (approved):

AaronErhardt:
Thanks!

PR #86, opened 2022-01-26
feat: Redesign of component architecture

mmstick:
## Proposal

This is a prototype for an alternative design that combines all of the model and widget traits into one `Component` trait that eliminates the dependency on parent type info. An alternative trait could be provided that simplifies component setup by combining the update and update_view methods into one.

## Component Initialization

There are a few stages that a component goes through when being created.

1. `Component::init()` creates a `Bridge` which contains public access to the `Root` widget
2. `Bridge::launch(Payload)` takes a `Payload` which contains the initial parameters necessary for creating the model and view from the `Component::dock()` static method. This spawns the event loop handling `Component::Input` messages and returns a `Fairing`.
3. The `Fairing` type contains three methods:
    -  `detach()` consumes it and returns a `Controller`, which has the receiver dropped
    - `forward(input.clone(), |output_event| callers_event)` transforms `Output` events  into an event compatible with the `input` sender. This method also returns a `Controller`
    - `connect_receiver(move |sender, event| {})` instead takes a mutable closure that's called on each received output, and also returns a `Controller`

## Examples

Included is a purposely-convoluted example which tries to utilize most of the features presented. Not included in the example is the forwarding method that a component can use to forward events from a sub-component to itself (or to an output). No macro included at this time.

## Thoughts

The current drawback is a hard dependency on tokio for sending commands to be processed by a tokio runtime, and by workers.


src/component/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+        params: Self::InitParams,
+        root: &Self::Root,
+        input: &mut Sender<Self::Input>,
+        output: &mut Sender<Self::Output>,
+    ) -> ComponentParts<Self, Self::Widgets>;
+
+    /// Initializes the component and attaches it to the thread-local GLib executor.
+    ///
+    /// The context's service remains alive for as long as the root widget is alive.
+    fn init(
```
AaronErhardt, line 52: My idea for init was that it is already the second step. Let's think about the following example:

There's a component that needs to be added to your widgets in the correct order (between two other widgets). This component also need to call `set_transient_for` and therefore has the parent window in the `InitParams`. The problem now is that you have to pass the parent window before you have created the component. This requires some awkward manual ordering to create the window first, pass it to the component and then initialize the window with all its children.

My suggestion would be to just use `init_root()` for `RawComponent` and then use finalize for the actual initialization. With this the widget of a component is available from the beginning and all the ordering issues should disappear.

AaronErhardt (commented):
Looks pretty good overall! I will have a closer look and do some testing this weekend.

mmstick:
I see. I'll get that done after breakfast. I have also just added a replacement for the worker trait with the last commit.

mmstick:
I have removed `async-trait` because it causes the `Component` trait to require itself to be `Send` just to call a static method from a background thread. Async methods simply return a pinned boxed future instead.

I have also added a `StatefulComponent` trait, which combines `update_view` into `update`, and reuses the same types. It currently requires launching a component with `launch_stateful(InitialParams)` rather than `launch(InitialParams)`.

It's no longer required for the programmer to manually set up a tokio runtime. A tokio runtime is managed internally that defaults to `1` worker thread and `512` max blocking threads. The runtime is launched when the first component or worker is spawned onto it. Two static global variables, `RELM_THREADS` and `RELM_BLOCKING_THREADS`, can be defined by the programmer before the first component is launched to change the defaults.

Added `spawn` and `spawn_blocking` to go along with `spawn_local`. `spawn_local` returns a `GLib::SourceID` that can be used to call `.remove()` on it to drop the future on GLib's executor. `spawn` and `spawn_blocking` return `tokio::JoinHandle<T>` which can be used to call `.abort()` on them.

AaronErhardt:
Sounds good. If you want you can set the target branch to "new-approach" so we can merge this and work on this branch until some testing was done and most examples are ported to the new version.

As I said, I will have a closer look this weekend or maybe already Friday afternoon (which should be Friday morning for you).

AaronErhardt (approved):
So, I've had a closer look at the code. Well done! Everything appears very clean to me.

I've also summarized the new concept so it's easier to reason about the design:

## Component initialization:

`Bridge (+Payload) -> Fairing -> Controller`

### Bridge:
- Gives you access to the root widget
- Can be converted into `Fairing` with a given `Payload`
- During the conversion, it spawns the runtime that handles incoming messages (and runs commands)

### Fairing:
- Either spawn a new runtime to forward messages from the component with a given sender
- Or connect a function that receives messages from the component and can forward the messages
- Or do nothing and convert itself into an controller

> Do these methods really need to consume the `Fairing`? 
> Maybe connecting multiple things might have a use-case.

### Controller:
- Gives you access to the root widget
- Allows you to send messages to the component
- *This is what you want to keep in your model*

I like the new design and also the analogy to rockets, but for someone like me who's not a native speaker words like fuselage, fairing or bridge are rather foreign in this context, even after watching a few streams from SpaceX. Maybe we can find another analogy or just words that are known more commonly.

And if you're ready to merge this, let me know by removing the wip status.

mmstick:
It's ready. Changing names to something else is fine, too.

mmstick:
Also, the fairing is consumed when handling the receiver because it has to detach the receiver since its a mpsc channel.

PR #88, opened 2022-01-28
WIP: Add fairing type that handles multiple receivers

AaronErhardt:
I thought that having multiple receiver functions might be useful. This is just a very quick implementation for now, but it already satisfies the compiler. 

I haven't checked if it's actually usable though.

@mmstick what do you think?

### DIFF
--- a/src/component/fairing.rs
+++ b/src/component/fairing.rs
@@ -20,11 +20,55 @@ pub struct Fairing<W, I, O> {
     pub receiver: Receiver<O>,
 }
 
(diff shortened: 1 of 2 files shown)


mmstick:
Similar could be achieved by using `tokio::sync::broadcast`. The receiver could either be part of a broadcast channel, or it could be generated after by consuming the original receiver and then sending that through a new broadcast channel. And I think there's also a potential use case for watch channels as well.

PR #91, opened 2022-01-30
Allow root widget to be any type

mmstick:
- Makes the `Bridge::attach_to` method available for `StatefulComponent` types
- Permits components to have any type as the root widget, instead of just GTK widgets
- Adds `Bridge::preflight` for pre-launch configuration for everything else

### DIFF
--- a/src/component/bridge.rs
+++ b/src/component/bridge.rs
@@ -16,6 +16,14 @@ pub struct Bridge<Component, Root> {
     pub(super) component: PhantomData<Component>,
 }
 
+impl<Component, Root> Bridge<Component, Root> {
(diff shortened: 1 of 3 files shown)


AaronErhardt:
I like this idea of these convenience methods, but I also think they hide the need of a component for `set_transient_for`. I think the better approach for components would be to take `Option<Window>` in the payload. This makes it clear that you should pass a window.

mmstick:
Then I will remove all but the hard requirement on the root type being a GTK widget.

AaronErhardt (approved):
Looks good to me. Is it ready to merge?

mmstick:
Yes, it's complete as is.

PR #93, opened 2022-01-31
Implemented FactoryView and FactoryListView for Leaflet

pentamassiv:
Added the implementations of the traits 'FactoryView' and 'FactoryListView' for adwaitas Leaflet type


AaronErhardt:
Thanks! :+1:

PR #95, opened 2022-01-31
feat!: Separate command outputs from component inputs

mmstick:
Requires #91 to be merged in advance. PR shall be rebased after merged.

Removes the `Send` restriction on `Component::Input`. Adds `update_cmd` methods for handling the new `Component::CommandOutput` types.


src/component/bridge.rs
```diff
@@ -124,3 +152,52 @@ impl<C: StatefulComponent> Bridge<C, C::Root> {
         }
     }
 }
+
+/// Messages being forwarded into the a component's main service.
+enum InnerMessage<Input, CmdOutput> {
+    ModelUpdate(Input),
+    CommandUpdate(CmdOutput),
+}
```
AaronErhardt, line 160: This is necessary because the model must be captured by only one closure, right? It might be nice to leave a note for anyone who reads the code so nobody wonders why it doesn't simply listen on two receivers.

AaronErhardt (approved):
Looks good to me.

src/component/bridge.rs
mmstick: Yea, I can add the note.

mmstick:
Now possible to externally borrow the model and widgets from a component. Borrowing with `controller.state.ro()` won't trigger an update of the view, but `controller.state.rw()` will cause `update_view()` to get called with a `Component`, and `update_notify` from a `StatelessComponent`.

src/component/state_watcher.rs
```diff
@@ -0,0 +1,27 @@
...
+#[derive(Debug)]
+pub struct StateWatcher<Component, Widgets> {
+    /// The models and widgets maintained by the component.
+    pub(super) state: RefCell<Fuselage<Component, Widgets>>,
+    pub(super) notifier: Rc<Notify>,
+}
+
+impl<Component, Widgets> StateWatcher<Component, Widgets> {
+    /// Borrows the model and view of a component.
+    pub fn ro(&self) -> Ref<'_, Fuselage<Component, Widgets>> {
```
AaronErhardt, line 18: I think `ro` and `rw` are a little to short. It's easy to guess if you're already familiar with Relm4 but for beginners it might be confusing. Something like `get` and `get_mut` would be easier to understand.

mmstick: Changed to `get`

PR #97, opened 2022-02-01
Simple component

AaronErhardt:
Adds a simpler component trait that implements the normal component trait. This should help reducing the complexity for simple components.

Requires #96  to be merged before.

@mmstick can you review?


mmstick (approved):
This is fine 👍

Issue #84, opened 2022-01-24
re-initialization of FactoryVec seems not to work


tronta:
In my Sudoku Solver example, I wanted to reinitialize/reset of a 
https://github.com/AaronErhardt/Relm4/pull/83/files#diff-d648c225a07e9a1e3276b116cdc0265e8f74c9ebe8d453db85e8920ab5aa4f0cR78

I tried it in two ways:
* assigning the FactoryVec a new vector `self.values = AppModel::empty().values;`
* clearing it before beforehand with `self.values.clear();`

The result was in both ways that new widgets were added, but the old ones not removed.

AaronErhardt:
Can you try again with the latest commits? The first solution won't work because I can't properly implement `Drop`for the `FactoryVec`, but the second one should now be fixed.

tronta:
```
              self.values.clear();
              self.values = AppModel::new().values;
```
did not work. It's still adding new widget on top the others.

AaronErhardt:
Yeah, you can't do the second line. When the factory is dropped the widgets are still alive. The only thing that works it calling `.clear()` and then re-render the factory in `view()`. As I said, I can't simply implement `Drop` so dropping the factory is almost always wrong.

AaronErhardt:
I've added this to the 0.5 milestone. I will try to update factories so they support `Drop` which will fix this problem.

PR #104, opened 2022-02-03
Fix FactoryVec state changes

V02460:
`FactoryVec` was acting weird and all over the place for me and it might be #84, too. I had some fun and modeled its fields as a state machine. Brings `push`, `pop`, and `clear` in line with the described model and fixes the indexing problems in the `generate` method. I think it is (more) correct now, but one could probably be smarter about `clear` and the use of `Vec::remove`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,5 +42,7 @@ macros = ["relm4-macros"]
 all = ["tokio-rt", "libadwaita", "macros"]
 
 [dev-dependencies]
+futures-channel = "0.3"
+futures-executor = "0.3"
 # Macros required to run tests.
 relm4-macros = { path = "relm4-macros" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,7 @@ libadwaita = ["adw"]
 tokio-rt = ["tokio", "async-trait"]
 macros = ["relm4-macros"]
 all = ["tokio-rt", "libadwaita", "macros"]
+
+[dev-dependencies]
+# Macros required to run tests.
+relm4-macros = { path = "relm4-macros" }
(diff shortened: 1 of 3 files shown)


AaronErhardt (approved):
Looks good! Thanks for the explanation in the code, that should make it a lot easier to understand.

Is it ready to merge?

V02460:
I pushed one small fix. Now it’s fine from my side. Works for me™ :stuck_out_tongue:

In all seriousness, though… I should add some tests.

AaronErhardt:
Yes, tests are definitely a good idea :)
It's hard to reason whether the code works or not.

V02460:
Here are some unit tests 🎁 I made two separate commits, one for the tests themself and another one for getting GTK to play nicely along with multi-threaded test running. I changed the `Cargo.toml` for two reasons:
1. I got advised by the gtk-rs guys to adapt their test running helper, so I did that.
2. To use relm4 macros in the implementation of the tests.

Cargo.toml
```diff
@@ -24,7 +24,9 @@ members = [
 adw = { version = "0.1", optional = true, package = "libadwaita" }
 async-trait = { version = "0.1", optional = true }
 fragile = "1.1"
+futures-channel = "0.3"
 futures-core = "0.3"
+futures-executor = "0.3"
```
AaronErhardt, line 29: Shouldn't those also be dev-depencencies?

AaronErhardt (commented):
Looks good. Actually, some of this testing infrastructure could eventually be exposed to help resolving https://github.com/AaronErhardt/Relm4/issues/78. But that's another topic.

Cargo.toml
V02460: Yes, pushed the change

AaronErhardt:
Thanks a lot!

PR #105, opened 2022-02-05
small imrpovement on calc-trainer UI

tronta:
a small change which makes the UI of the calc-trainer a little more appealing.


AaronErhardt:
Thanks!

PR #108, opened 2022-02-12
New approach macro

### LINKED ISSUE #99: Update macros for the new component trait
- [x] Agree on new macro syntax
- [x] Implement the macro

AaronErhardt:
- Add a `component` macro to replace the `widget` macro.
- Reintroduce the ["simple" example](https://github.com/AaronErhardt/Relm4/blob/new-approach-macro/examples/simple.rs)
- Eventually fixes #99


AaronErhardt (commented):
I think it's good enough to merge this now. Yet of course, there is still room for improvement.

PR #111, opened 2022-02-18
0.5: Graceful Shutdown

mmstick:
The current component traits have a big problem with handling graceful shutdowns. This change adds a new method to the component traits — `shutdown()` — which allows a component to get one last chance to perform a cleanup routine before its terminated. It also hands a broadcast receiver to each spawned command so that they can be notified of a component shutdown.

### DIFF
--- a/examples/progress.rs
+++ b/examples/progress.rs
@@ -105,7 +105,7 @@ impl Component for App {
                 append: button = &gtk::Button {
                     set_label: "Compute",
                     connect_clicked(input) => move |_| {
-                        let _ = input.send(Input::Compute);
+                        input.send(Input::Compute);
                     }
                 }
             }
@@ -186,12 +186,12 @@ impl Component for App {
                 let mut progress = 0.0;
 
                 while progress < 1.0 {
-                    let _ = out.send(CmdOut::Progress(progress));
+                    out.send(CmdOut::Progress(progress));
                     progress += 0.1;
                     tokio::time::sleep(std::time::Duration::from_millis(333)).await;
                 }
 
-                let _ = out.send(CmdOut::Finished(Ok("42".into())));
+                out.send(CmdOut::Finished(Ok("42".into())));
             })
             // Perform task until a shutdown interrupts it
             .drop_on_shutdown()
--- a/examples/progress.rs
+++ b/examples/progress.rs
@@ -194,7 +194,7 @@ impl Component for App {
                 let _ = out.send(CmdOut::Finished(Ok("42".into())));
             })
             // Perform task until a shutdown interrupts it
-            .wait_then_drop()
+            .drop_on_shutdown()
             // Wrap into a `Pin<Box<Future>>` for return
             .boxed()
     }
--- a/examples/progress.rs
+++ b/examples/progress.rs
@@ -1,0 +1,201 @@
+// Copyright 2022 System76 <info@system76.com>
+// SPDX-License-Identifier: MIT or Apache-2.0
+
+//! A component which allows the caller to define what options ae in its list.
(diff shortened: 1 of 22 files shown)


examples/settings_list.rs
```diff
@@ -238,13 +239,20 @@ impl Component for SettingsListModel {
...
+                    SettingsListCommand::Reload => {
+                        tokio::time::sleep(std::time::Duration::from_secs(2)).await;
+                        Some(SettingsListCmdOutput::Reload)
+                    }
                 }
+            };
+
+            futures::select! {
+                _ = shutdown.recv().fuse() => None,
+                response = task.fuse() => response,
```
AaronErhardt, line 255: Can we make this easier? I think most components would just like to drop the future when the shutdown signal is received.

AaronErhardt (approved):
Looks good!

mmstick:
How about this:

```rs
fn command(message: Self::Command, shutdown: ShutdownReceiver) -> CommandFuture<Self::CommandOutput> {
    shutdown
        // Performs this operation until a shutdown is triggered
        .register(async move {
            match message {
                SettingsListCommand::Reload => {
                    tokio::time::sleep(std::time::Duration::from_secs(2)).await;
                    Some(SettingsListCmdOutput::Reload)
                }
            }
        })
        // Returns this value if a shutdown was triggered.
        .unwrap_or(async { None })
        // Wrap into a `Pin<Box<Future>>` for return
        .boxed()
}
```

mmstick:
I'm thinking about changing commands to have access to the command output sender, so it's possible to have progress and log messages reported back, instead of only getting a chance to send back info at the end.

mmstick:
Yeah I'm going to add progress examples soon, among rewriting existing examples. It looks like I'll have more time for this in the coming week.

examples/progress.rs
```diff
@@ -0,0 +1,203 @@
...
+
+use futures::FutureExt;
+use gtk::prelude::*;
+use relm4::*;
+
+fn main() {
+    gtk::Application::builder()
+        .application_id("org.relm4.ProgressExample")
+        .launch(|_app, window| {
+            // Intiialize a component's root widget
```
AaronErhardt, line 23: Typo

examples/progress.rs
```diff
@@ -0,0 +1,203 @@
...
+//! component, which then issues to reload the widgets again.
+
+use futures::FutureExt;
+use gtk::prelude::*;
+use relm4::*;
+
+fn main() {
+    gtk::Application::builder()
+        .application_id("org.relm4.ProgressExample")
+        .launch(|_app, window| {
```
AaronErhardt, line 22: I prefer using `RelmApp` for startup as it reduces the boilerplate code. Yet of course `ApplicationBuilder` is fine, too.

examples/progress.rs
```diff
@@ -0,0 +1,203 @@
...
+                    connect_clicked(input) => move |_| {
+                        input.send(Input::Compute);
+                    }
+                }
+            }
+        }
+
+        root.append(&container);
+
+        progress.hide();
```
AaronErhardt, line 116: `set_visible` in the view! macro would be shorter

AaronErhardt (commented):
Good job overall!

examples/progress.rs
mmstick: Indeed. And it's not necessary to use append anymore.

AaronErhardt (approved):
I suggest to merge this soon as it's already a quite large PR. We can address the other issues in separate PRs.

mmstick:
Yeah, this is ready

PR #116, opened 2022-02-23
Add a few improvements to reduce the amount of clone in user code

cofee-on-the-desk:
I'm experimenting with the new components in my project. I often find myself cloning stuff, in two situations: 

### 1) I'd like to attach a window or a dialog to a window (here's an example from my code):

```rust
// ... we have a 'window' variable defined somewhere before
let properties = {
            let window = window.clone();
            PropertyWindow::init()
                .update_root(move |w| w.set_transient_for(Some(&window)))
                .launch(())
                .detach()
            };
```

I have to clone the `window` here, as the closure takes ownership of it's arguments.

The new method `transient_to` can help remove this clone. I use 'to' here for the analogy to an existing method, `attach_to`.

```rust
let properties = PropertyWindow::init()
            .transient_to(&window)
            .launch(())
            .detach();
```

### 2) I'd like to forward a message:

```rust
let rules = {
            let sender = sender.clone();
            RuleViewModel::init()
                .launch(data.current_dir_rules())
                .forward(sender, move |msg| {
                    // .. forward stuff here
                });
        };
```

Forward requires ownership of `sender`, so we have to copy it. I changed it now to use a reference:

```rust
let rules = RuleViewModel::init()
            .launch(data.current_dir_rules())
            .forward(&sender, move |msg| {
                // .. forward stuff here
            });
```

P.S. This is my first ever pull request - hope I did it correctly =)

### DIFF
--- a/src/component/builder/mod.rs
+++ b/src/component/builder/mod.rs
@@ -37,7 +37,7 @@ impl<Component, Root: AsRef<gtk::Widget>> ComponentBuilder<Component, Root> {
 
 impl<Component, Root: AsRef<gtk::Window>> ComponentBuilder<Component, Root> {
     /// Set the component's root widget transient for a given window.
-    pub fn transient_to(self, window: impl AsRef<gtk::Window>) -> Self {
+    pub fn transient_for(self, window: impl AsRef<gtk::Window>) -> Self {
         self.root.as_ref().set_transient_for(Some(window.as_ref()));
 
         self
--- a/src/component/builder/mod.rs
+++ b/src/component/builder/mod.rs
@@ -6,6 +6,7 @@ mod elm_like;
 mod stateful;
 
 use crate::RelmContainerExt;
+use gtk::prelude::GtkWindowExt;
 use std::marker::PhantomData;
 
 /// A component that is ready for docking and launch.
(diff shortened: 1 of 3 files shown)


AaronErhardt:
Thanks for sending a PR! I like the addition of `transient_to` (although it should probably be called `transient_for`).

I wonder though, wouldn't this work for the first example?

```rust
        let properties = {
            PropertyWindow::init()
                .update_root(|w| w.set_transient_for(Some(&window)))
                .launch(())
                .detach()
            };
```

I don't see why a closure with `move` would be needed in this case but I could be wrong.

And for the second, couldn't you write this?

```rust
        let rules = {
            RuleViewModel::init()
                .launch(data.current_dir_rules())
                .forward(sender.clone(), move |msg| {
                    // .. forward stuff here
                });
        };
```

Regardless of my concerns, you did really well with your first PR!
At least `transient_for` looks like a nice addition to me.

@mmstick what do you think?

mmstick (approved):
Looks good to me

cofee-on-the-desk:
@AaronErhardt here's an error that occurs when I try to not use move:

```rust
error[E0373]: closure may outlive the current function, but it borrows `window`, which is owned by the current function
  --> src/main.rs:90:26
   |
90 |             .update_root(|w| w.set_transient_for(Some(&window)))
   |                          ^^^                           ------ `window` is borrowed here
   |                          |
   |                          may outlive borrowed value `window`
   |
note: function requires argument type to outlive `'static`
  --> src/main.rs:89:26
   |
89 |           let properties = PropertyWindow::init()
   |  __________________________^
90 | |             .update_root(|w| w.set_transient_for(Some(&window)))
   | |________________________________________________________________^
help: to force the closure to take ownership of `window` (and any other referenced variables), use the `move` keyword
   |
90 |             .update_root(move |w| w.set_transient_for(Some(&window)))
   |                          ++++
```

The second example - yeah, you could write it like this, I think this was just the habit of copying senders into closures =)

AaronErhardt:
Right, because the closure must be static the reference alone isn't enough. I wonder if that can be changed to a less restrictive lifetime...

Anyway, apart from renaming to `transient_for` it looks good and you're probably right that cloning less makes the API cleaner.

AaronErhardt (approved):
Thanks!

PR #118, opened 2022-02-24
Add `RelmRemovableExt::remove_all` and `RelmWidgetExt::iter_children_reverse`

cofee-on-the-desk:
Introduce `RelmRemovableExt::remove_all` method, which would allow to remove all children from a widget. I think it could be useful.

I've had to change the trait implementation a bit, because otherwise you would have to specify the generic type parameter every time you use this method (like `RelmRemovableExt::<gtk::Widget>::remove_all(&container)`). The static type checking remains though.

If the implementation change is not acceptable for some reason, I can drop it and instead introduce a separate trait for this method.

### DIFF
--- a/src/extensions/mod.rs
+++ b/src/extensions/mod.rs
@@ -40,8 +40,11 @@ impl ApplicationBuilderExt for gtk::builders::ApplicationBuilder {
 
 /// Additional methods for `gtk::Widget`
 pub trait RelmWidgetExt {
-    /// Iterates across the child widgets of a widget
+    /// Iterates across the child widgets of a widget.
     fn iter_children(&self) -> Box<dyn Iterator<Item = gtk::Widget>>;
+
+    /// Iterates across the child widgets of a widget, in reverse order.
+    fn iter_children_reverse(&self) -> Box<dyn Iterator<Item = gtk::Widget>>;
 
     /// Iterates children of a widget with a closure.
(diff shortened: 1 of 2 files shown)


AaronErhardt:
I like the idea and the code looks solid, but to be honest, there can be some quite unexpected things happening when removing GTK widgets. There are some widgets like `HeaderBar` that consist of boxes internally where the actual children are attached (I've done this wrong in the past, too: https://github.com/AaronErhardt/Relm4/issues/103#issuecomment-1029534362). So to be sure that the code is actually correct we will need tests, at least for the types that aren't obvious. 

The infrastructure for the tests could be inspired by this: https://github.com/gtk-rs/gtk4-rs/blob/master/gtk4/src/lib.rs#L61

Let me know if you need any help.

cofee-on-the-desk:
Could you write the test scheduler yourself? You are way more familiar with the gtk-rs/relm4 codebase than I am.

AaronErhardt:
I'm planning to fix this together with the team at gtk-rs: https://github.com/gtk-rs/gtk4-rs/issues/951

I hope it doesn't take long to setup the infrastructure, otherwise we can also merge this without test and add them later.

cofee-on-the-desk:
Cool! If the code is okay, we can merge now. I'll add the tests as soon as the support for them lands.

AaronErhardt:
Ok then let's merge this now. Thanks!

PR #119, opened 2022-02-24
implement factory traits for libpanel Paned

euclio:
libpanel is still pretty unstable, so I'm not sure if we want to add any additional stability warnings here (we might want to upgrade the library in a non-semver compatible way).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ members = [
 
 [dependencies]
 adw = { version = "0.1", optional = true, package = "libadwaita" }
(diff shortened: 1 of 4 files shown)


AaronErhardt:
I guess users of libpanel will be aware of potential semver issue. Yet, you could at least add a re-export of libpanel like it is the case with gtk and libadwaita, so you don't need to explicitly specify it as dependency.

Also, did you run `cargo fmt`?

Otherwise it looks good to me.

euclio:
Ran rustfmt and added the re-export.

AaronErhardt:
Thanks!

PR #121, opened 2022-02-25
refactor!: Merge `StatefulComponent` with `Component` trait

mmstick:
Adds two new methods to the `Component` trait to support both modes: `update_with_view` and `update_cmd_with_view`. By default, these methods are already implemented and will simply invoke `update` + `update_view` and `update_cmd` + `update_view` by default.

This also made it possible to simplify most of the types to accept a single generic `Component` type.


AaronErhardt (approved):
Looks good. It's remarkable how few changes this required in the examples.

PR #122, opened 2022-02-26
Impl `remove` and `swap_remove` for FactoryVec.

songww:


AaronErhardt:
Code looks good. Could you add some test, though? The tests are located at the bottom of the file, just adding two small test cases for both methods would be great. Factory code is quite fragile and with tests we can ensure that no regressions happen.

PR #126, opened 2022-03-02
use associated consts for action names

euclio:
Minor breaking change to the action traits, but it makes implementing
them manually a bit less verbose. Maybe someday it will let us compute
the full action name at compile-time, too.

### DIFF
--- a/src/actions/mod.rs
+++ b/src/actions/mod.rs
@@ -17,9 +17,7 @@ macro_rules! new_action_group {
         $vis struct $ty;
 
         impl relm4::actions::ActionGroupName for $ty {
-            fn group_name() -> &'static str {
-                $name
-            }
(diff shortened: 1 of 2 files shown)


AaronErhardt (approved):
Thanks for the PR! I wasn't aware of associated consts, but they definitely make a lot of sense in this case.

However, I'm not sure whether it makes sense to introduce a breaking change when there will be soon a beta release for 0.5...

euclio:
Yep, it's minor and unnecessary so feel free to save this for a batch of other breaking changes.

AaronErhardt:
It would be great if you could send a new PR to target the "new-approach" branch. I will soon merge it into main and release a beta version of 0.5.

euclio:
Sounds good. I rebased and changed the destination of this PR.

PR #127, opened 2022-03-08
Drop general extensions in favor of widget-specific extensions

cofee-on-the-desk:
1) Remove `RelmRemovableExt::remove_all`. It relies on widget internal structure and may break between releases, which would be hard to maintain.
2) Remove `RelmWidgetExt::iter_children`, `RelmWidgetExt::iter_children_reverse` and `RelmWidgetExt::for_each_children`. Those aren't really reliable and can yeild surprising results on some widgets, like `gtk::HeaderBar` (see [link](https://github.com/AaronErhardt/Relm4/issues/103#issuecomment-1029534362)).
3) Instead, introduce widget-specific extensions: `RelmBoxExt`, `RelmListBoxExt`, `RelmFlowBoxExt` and `RelmGridExt`. Add tests for those extensions.

GTK4 encourages widget-specific implementations, instead of generalizations. See [link](https://discourse.gnome.org/t/gtk4-widget-container-api-model/7672).

### DIFF
--- a/src/extensions/mod.rs
+++ b/src/extensions/mod.rs
@@ -8,7 +8,7 @@ mod tests;
 #[allow(unreachable_pub)]
 pub use self::container::RelmContainerExt;
 #[allow(unreachable_pub)]
-pub use self::removable::RelmRemovableExt;
+pub use self::removable::{RelmRemovableExt, RelmRemoveAllExt};
 #[allow(unreachable_pub)]
 pub use self::set_child::RelmSetChildExt;
 
@@ -63,49 +63,10 @@ impl<T: gtk::glib::IsA<gtk::Widget>> RelmWidgetExt for T {
     }
 }
 
-fn iter_children(widget: &gtk::Widget) -> impl Iterator<Item = gtk::Widget> {
-    let mut widget = widget.first_child();
-
-    std::iter::from_fn(move || {
-        if let Some(child) = widget.take() {
-            widget = child.next_sibling();
-            return Some(child);
-        }
-
-        None
-    })
-}
-
-/// Additional methods for `gtk::Box`.
-pub trait RelmBoxExt {
-    /// Returns all children of the box.
-    fn children(&self) -> Vec<gtk::Widget>;
-
-    /// Remove all children from the box.
-    fn remove_all(&self);
-}
-
-impl RelmBoxExt for gtk::Box {
-    fn children(&self) -> Vec<gtk::Widget> {
-        iter_children(self.as_ref()).collect()
-    }
-    fn remove_all(&self) {
-        while let Some(child) = self.last_child() {
-            self.remove(&child);
-        }
-    }
-}
-
 /// Additional methods for `gtk::ListBox`.
 pub trait RelmListBoxExt {
-    /// Returns all rows of the listbox.
-    fn rows(&self) -> Vec<gtk::ListBoxRow>;
-
     /// Get the index of a widget attached to a listbox.
     fn index_of_child(&self, widget: &impl AsRef<gtk::Widget>) -> Option<i32>;
-
-    /// Remove all children from listbox.
-    fn remove_all(&self);
 
     /// Remove the row of a child attached a listbox.
     fn remove_row_of_child(&self, widget: &impl AsRef<gtk::Widget>);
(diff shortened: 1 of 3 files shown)


src/extensions/mod.rs
```diff
@@ -99,6 +115,15 @@ pub trait RelmListBoxExt {
...
+            .map(|widget| {
+                widget
+                    .downcast::<gtk::ListBoxRow>()
+                    .expect("The child of `ListBox` is not a `ListBoxRow`.")
+            })
+            .collect()
+    }
     fn index_of_child(&self, widget: &impl AsRef<gtk::Widget>) -> Option<i32> {
         if let Some(row) = self.row_of_child(widget) {
             return Some(row.index());
```
AaronErhardt, line 129: Can't you do something like `self.row_of_child(widget).map(|row| row.index())` ?

src/extensions/mod.rs
```diff
@@ -86,11 +63,50 @@ impl<T: gtk::glib::IsA<gtk::Widget>> RelmWidgetExt for T {
...
+    })
+}
+
+/// Additional methods for `gtk::Box`.
+pub trait RelmBoxExt {
+    /// Returns all children of the box.
+    fn children(&self) -> Vec<gtk::Widget>;
+
+    /// Remove all children from the box.
+    fn remove_all(&self);
```
AaronErhardt, line 85: `remove_all` appears more often and similar methods like `children`, too.

I don't want to put everything in one trait, yet where some widgets share the same functionality, a shared trait would make sense.

Maybe something like this:

```rust
trait RelmRemoveAll {
    fn remove_all(&self);
}

trait RelmChildren {
    type Child;
    fn children(&self) -> Vec<Child>;
}
```

Also, ideally every `children` method should return an iterator instead of an vector.

AaronErhardt (changes requested):
Good work! Looks really good overall.
In the long term, some of this functionality could potentially even be upstreamed to gtk-rs.

src/extensions/mod.rs
cofee-on-the-desk: We can have traits if we keep them as simple as possible.

Making `children` an iterator is a good idea.

src/extensions/mod.rs
cofee-on-the-desk: Sure.

src/extensions/mod.rs
cofee-on-the-desk: Take a look at the changes. I've added the traits as you have suggested. Also it's worth noting that `iter_children` method returns a `ChildrenIterator<Self>`, which itself implements `DoubleEndedIterator` and makes it possible to iterate from both sides of the container.

src/extensions/mod.rs
```diff
@@ -130,28 +104,96 @@ impl RelmListBoxExt for gtk::ListBox {
...
+                    child
+                        .downcast::<T::Child>()
+                        .expect("The type of children does not match.")
+                });
+                return Some(start);
+            }
         }
 
         None
-    })
```
AaronErhardt, line 143: If you use 
```rust
if self.done {
    None
else {
    ....
    Some(start)
}
```

You can get rid of the `return` and make the program flow more obvious. Same for `DoubleEndedIterator`.

src/extensions/removable.rs
```diff
@@ -1,193 +1,40 @@
-use crate::{RelmSetChildExt, RelmWidgetExt};
+use crate::RelmSetChildExt;
 use gtk::prelude::*;
 
 /// Widget types which can have widgets removed from them.
-pub trait RelmRemovableExt {
-    /// Type of children of the container.
-    type Child;
+pub trait RelmRemovableExt<W> {
```
AaronErhardt, line 5: Is there a specific reason to use a generic parameter `W` instead of an associated type `Child`? I can image some cases where this might require some ugly type annotations because Rust can't know the generic type parameter.

src/extensions/tests.rs
```diff
@@ -0,0 +1,136 @@
...
+) -> bool {
+    match (first, second) {
+        (Some(first), Some(second)) => first.as_ref() == second.as_ref(),
+        (None, None) => true,
+        _ => false,
+    }
+}
+
+#[gtk::test]
+fn box_ext() {
```
AaronErhardt, line 29: A name like `box_extension_trait_test` would be more fitting now that `BoxExt` was removed. Same for grid and flow_box.

AaronErhardt (changes requested):
Excellent work! This is shaping up nicely and I really like the clean implementations for `RelmIterChildrenExt` and the `DoubleEndedIterator` implementation. I couldn't have done this better :)

There are just a few small things, I'd like to sort out before merging it.

src/extensions/removable.rs
cofee-on-the-desk: No, there isn't. I'll change it then.

src/extensions/tests.rs
cofee-on-the-desk: `box_extension_traits` should be fine.

src/extensions/removable.rs
cofee-on-the-desk: There might be a name conflict though if `RelmRemovableExt` and `RelmIterChildrenExt` are in the same scope.

AaronErhardt: That's true. Probably, all traits that have `RelmIterChildrenExt` also implement `RelmRemovableExt`, or am I wrong. I think you could use `trait RelmIterChildrenExt: RelmRemovableExt { ... }` to use the `Child` type from one trait in the other trait, too. Or you just rename one to `RemoveChild` or `IterChild`.

cofee-on-the-desk:
Done. Also I've noticed incorrect behavior of `ChildrenIterator` in some cases - it's fixed now. If you have any Ideas how to simplify the implementation - I'd love to hear them.

src/extensions/mod.rs
```diff
@@ -130,28 +104,121 @@ impl RelmListBoxExt for gtk::ListBox {
...
+            });
+            if self.start == self.end {
+                if self.init_end {
+                    return None;
+                } else {
+                    self.init_end = true;
+                    return self.start.clone();
+                }
+            }
+            return self.start.clone();
```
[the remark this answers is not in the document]
AaronErhardt: Wait a minute, I think I've overlooked a case where you use `next_back()`.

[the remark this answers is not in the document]
AaronErhardt: Never mind, I actually think it's correct to use what I suggested, if `DoubleEndedIterator` does the same thing in reverse. Just had to go through some cases in my mind, but couldn't find one that fails.

AaronErhardt (approved):
Well done, excellent work! I think we can merge this now.

PR #129, opened 2022-03-11
Fix warning for `inline_css`, adjust API

cofee-on-the-desk:
Fix warning that occurs when you use `inline_css` without `;` at the end:

```rust
(process:108104): Gtk-WARNING **: 13:41:29.363: Theme parsing error: <data>:1:3-13: Expected ';' at end of block
```

A small API change - the method now requires `&str` instead of `&[u8]`.


AaronErhardt (approved):
Thanks!

PR #135, opened 2022-03-22
Rename `RelmRemovableExt` to `RelmRemoveExt`

cofee-on-the-desk:
A small fix to make the trait name more intuitive.


AaronErhardt:
Indeed, "removable" doesn't make sense for a container. Thanks for the PR!

Issue #101, opened 2022-02-02
Update relm4-components for the new component trait


AaronErhardt:
It'd be nice to support relm4-components in 0.5 from day one. These are the currently available components:

- [ ] open_button
- [ ] alert
- [ ] open_dialog
- [ ] save_dialog

And of course, more reusable components are always welcome!

cofee-on-the-desk:
I'd like to work on this issue.

AaronErhardt:
Awesome! Let me know if you need any help.

euclio:
How are the components expected to access their parent? By passing it through `InitParams`?

AaronErhardt:
Yes, the new component design makes them independent from their parent unless they require specific `InitParams` to access the parent. You can also forward the `Output` of the child to the `Input` of the parent.

PR #141, opened 2022-03-27
Port file dialogs to 0.5

cofee-on-the-desk:
Related to #101.

# Changes

- Port `OpenDialog` and `SaveDialog` to the next version or relm4.
- ~~Dialogs use a non-native file chooser.~~
    -  See #143 
- Added a new example, featuring a simple text editor.
- Added Default settings for both dialogs.
- Button labels now use `String` instead of `&'static str`, which allows creating them dynamically at runtime.  
- Support for selecting folders.
- Support for multi selection.

### DIFF
--- a/examples/relm4-components/file_dialogs.rs
+++ b/examples/relm4-components/file_dialogs.rs
@@ -95,7 +95,9 @@ impl SimpleComponent for App {
             .launch(OpenDialogSettings::default())
             .forward(sender.input_sender(), |response| match response {
                 OpenDialogResponse::Accept(path) => Input::OpenResponse(path),
-                _ => Input::ShowMessage(String::from("File opening was cancelled")),
+                OpenDialogResponse::Cancel => {
+                    Input::ShowMessage(String::from("File opening was cancelled"))
+                }
             });
 
         let save_dialog = SaveDialog::builder()
--- a/examples/relm4-components/file_dialogs.rs
+++ b/examples/relm4-components/file_dialogs.rs
@@ -1,14 +1,10 @@
-use gtk::prelude::{
-    ButtonExt, DialogExt, GtkWindowExt, OrientableExt, TextBufferExt, TextViewExt, WidgetExt,
-};
+use gtk::prelude::*;
 use relm4::{
     gtk, Component, ComponentController, ComponentParts, ComponentSender, Controller, RelmApp,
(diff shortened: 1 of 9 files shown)


euclio:
> Dialogs now use a non-native file chooser (reason: limitations on the component Root type).

We should open an issue to fix this generally, since this is a regression from the old method.

mmstick:
You might want to `git checkout next`, `git pull origin next`, `git checkout -b rebase`, `git cherry-pick {{SHA}}` from this branch, then `git branch -D next-components-file-dialogs` and `git checkout -b next-components-file-dialogs`. Perfect rebase with no rebase or merge commits.

cofee-on-the-desk:
Okay, thanks. Still trying to figure out how to work with git.

cofee-on-the-desk:
Almost finished. The API is a subject of change but the functionality should be solid.

The only thing left is to use native dialogs, and that should be fixed with #152.

examples/relm4-components/file_dialogs.rs
```diff
@@ -91,15 +91,15 @@ impl SimpleComponent for App {
         sender: &ComponentSender<Self>,
     ) -> ComponentParts<Self> {
         let open_dialog = OpenDialog::builder()
-            .transient_for(root)
+            .update_root(|dialog| dialog.set_transient_for(Some(root)))
```
AaronErhardt, line 94: This change is rather confusing regarding the API. I think adding a method like `transient_for_native` could make sense. What do you think?

cofee-on-the-desk: Sure, I'll add it. Too bad we can't implement both native and non-native dialogs in one method.

cofee-on-the-desk:
There's two components now: `OpenDialog`, which selects a single file, and `OpenDialogMulti`, which selects multiple.

This change allows for easier pattern matching on output - now you don't have to consider separate cases for `Accept(PathBuf)` and `AcceptMultiple(Vec<PathBuf>)`.

euclio:
What's the status of this PR?

cofee-on-the-desk:
@euclio I still have to adapt this PR to use the macro rework. But for now let's focus on getting `relm4` 0.5 beta released. After that we can continue our work on `relm4-components`.

AaronErhardt (approved):
Looks good, thanks!

PR #142, opened 2022-03-27
Builder

### LINKED ISSUE #140: new approach: `init` should not be a method on `Component`
The current design of the `Component` trait expects users to initialize a `ComponentBuilder` by calling `init()` on their `Model`. This feels weird to me, since reading `YourModel::init()` looks like it creates an instance of `YourModel`, not `ComponentBuilder`. Also, having it be a trait method allows users to override it in their `Component` impl, which is never something they should do.

Instead, why not define a method `ComponentBuilder::new()` that constructs a new builder, with the component type specified as a type parameter? This makes it clearer that a component is being constructed, and type inference will usually allow the model type to be inferred anyways. Another benefit of moving the method is that it would free up the valuable name `init` to be used instead of `init_parts`.

mmstick:
- Adds `ComponentBuilder::<C>::new()` as an alternative to `Component::init`
- Replaces `Component::init` with `Component::builder`
- Renames `Component::init_parts` with `Component::init`

Closes #140

### DIFF
--- a/examples/factory.rs
+++ b/examples/factory.rs
@@ -218,7 +218,7 @@ impl SimpleComponent for AppModel {
     }
 
     // Initialize the UI.
-    fn init_parts(
+    fn init(
         counter: Self::InitParams,
         root: &Self::Root,
         input: &Sender<Self::Input>,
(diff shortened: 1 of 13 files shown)


AaronErhardt:
Looks good, thanks!

PR #146, opened 2022-03-28
Add `SimpleComponent::builder`

cofee-on-the-desk:
Add `builder` to `SimpleComponent` so you don't have to import `Component` in the scope.


cofee-on-the-desk:
Actually there is a side effect: if you import both `Component` and `SimpleComponent` in the scope, to use `builder` you would have to specify the trait explicitly:

```rust
<App as Component>::builder()
    ...
```

This isn't a good tradeoff, so let's close this.

mmstick:
Perhaps we need a prelude module with all the traits.

PR #147, opened 2022-03-28
refactor!: Combine senders into ComponentSender type

mmstick:
This is an experiment to reduce boilerplate. Feedback would be appreciated if this seems like a good idea or not.

This will combine the two `input: Sender<Self::Input>, output: Sender<Self::Output>` parameters into a single `sender: ComponentSender<Self>`. This new sender type contains the two senders and also implements `fn input(&self, message: Self::Input)` and `fn output(&self, message: Self::Output)` for convenience. The `ComponentSender` is a type alias for an `Rc<ComponentSenderInner>` so it is clonable across signal callbacks.

This could also be extended to contain a command sender so that we could combine the `SimpleComponent` and `Component` traits back together.

### DIFF
--- a/relm4/src/component/sender.rs
+++ b/relm4/src/component/sender.rs
@@ -1,0 +1,40 @@
+// Copyright 2022 System76 <info@system76.com>
+// SPDX-License-Identifier: MIT or Apache-2.0
+
+use crate::{Component, Sender};
+use std::sync::Arc;
+
+/// Contain senders used by the component.
+pub type ComponentSender<C> = Arc<ComponentSenderInner<C>>;
+
+/// Contains senders used by the component.
+#[derive(Debug)]
+pub struct ComponentSenderInner<C: Component> {
+    /// Emits component inputs
+    pub input: Sender<C::Input>,
+
+    /// Emits component outputs
+    pub output: Sender<C::Output>,
+}
+
+impl<C: Component> ComponentSenderInner<C> {
+    /// Emit an input to the component.
+    pub fn input(&self, message: C::Input) {
+        self.input.send(message);
+    }
+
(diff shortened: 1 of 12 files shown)


relm4/src/component/sender.rs
```diff
@@ -0,0 +1,30 @@
+// Copyright 2022 System76 <info@system76.com>
+// SPDX-License-Identifier: MIT or Apache-2.0
+
+use crate::{Component, Sender};
+use std::rc::Rc;
+
+/// Contain senders used by the component.
+pub type ComponentSender<C> = Rc<ComponentSenderInner<C>>;
```
AaronErhardt, line 8: I don't think `Rc` is a good choice here, I'd use `Arc` instead. You might want to send messages from your own background thread or use commands as actors in a threaded runtime.

relm4/src/component/sender.rs
```diff
@@ -0,0 +1,30 @@
...
+    /// Emit an input to the component.
+    pub fn input(&self, message: C::Input) {
+        self.input.send(message);
+    }
+
+    /// Emit an output to the component.
+    pub fn output(&self, message: C::Output) {
+        self.output.send(message);
+    }
+}
```
AaronErhardt, line 30: Maybe some methods like `input_sender` and `output_sender` would be nice to access the senders.

AaronErhardt (changes requested):
I like the idea. It definitely helps to simplify the method definitions.

relm4/src/component/sender.rs
mmstick: Sure thing

AaronErhardt (approved):

PR #152, opened 2022-03-29
feat!: Configurable component stopping

mmstick:
- Pro: Removes the `OnDestroy` trait and `Root: OnDestroy` restriction.
- Pro: Can kill the component externally from the controller
- Pro: Can kill the component through any means necessary
- Con: Requires that the kill signal be attached by the component author.
    - Comes with `sender.stop_with_widget()` and `sender.stop_with_native_dialog()` to ease that burden.

### DIFF
--- a/examples/factory.rs
+++ b/examples/factory.rs
@@ -223,6 +223,9 @@ impl SimpleComponent for AppModel {
         root: &Self::Root,
         sender: &ComponentSender<Self>,
     ) -> ComponentParts<Self> {
+        // Stop the event loop when the root widget is destroyed.
+        sender.stop_with_widget(root);
+
         // Insert the macro codegen here
         let widgets = view_output!();
 
--- a/examples/grid_factory.rs
+++ b/examples/grid_factory.rs
@@ -237,6 +237,9 @@ impl SimpleComponent for AppModel {
         root: &Self::Root,
         sender: &ComponentSender<Self>,
     ) -> ComponentParts<Self> {
+        // Stop the event loop when the root widget is destroyed.
+        sender.stop_with_widget(root);
+
         // Insert the macro codegen here
         let widgets = view_output!();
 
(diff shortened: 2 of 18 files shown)


mmstick:
I have an idea for automatically connecting the destroy with the Builder.

mmstick:
Nevermind. It doesn't work.

AaronErhardt:
That's definitely an improvement over the status quo but maybe we can still use `OnDestroy` but implement it through `ObjectExt` which would work for `NativeDialog` as well.

It seems like this mechanism is actually working, it's very similar to what we have with `WidgetExt::connect_destroy`:

```rust
struct Dropper;

impl Drop for Dropper   {
    fn drop(&mut self) {
        // Actually drop the runtime here
        println!("DROPPED!");
   }
}

fn register_dropper(widget: impl ObjectExt) {
    let dropper = Dropper;
    let closure = gtk::glib::Closure::new(move |_| {
        // Move the "dropper"
        let _ = dropper;
        None
    });
    widget.watch_closure(&closure);
}
```

AaronErhardt:
I think we should decide how to address this soon. There are three possible routes: 

1. Keep `OnDestroy` but only require `ObjectExt` so basically all widgets even native ones are supported
2. Use additional methods to bind the lifetime to the widget or the controller from **inside** the component (like this PR)
3. Use additional methods to bind the lifetime to the widget or the controller from **outside** of the component (e. g. when after initialization)

Ideally, 2 and 3 should have a good default that doesn't require any code and does something reasonable (for example bind the lifetime of the runtime to the controller). Otherwise it'd be easy to forget.

Also, I think "Pro: Can kill the component externally from the controller" is something we can implement either way.

mmstick:
Sorry for the delay. I'll look into this soon.

AaronErhardt:
@mmstick FYI `ObjectExt` now has a method `add_weak_ref_notify` thanks to https://github.com/gtk-rs/gtk-rs-core/pull/666, which should work well as a replacement for `connect_destroy` and allows us to drop the requirement for `WidgetExt`.

mmstick:
Sounds good

AaronErhardt:
Superseded by #348

PR #153, opened 2022-03-30
run CI on all branches

euclio:
This should be ported to the `next` branch as well, so that we can get CI runs on PRs to that branch. (The build is currently broken on that branch, but it will be helpful to see if there's any unintended breakage).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,10 +1,6 @@
 name: CI
 
-on:
-  push:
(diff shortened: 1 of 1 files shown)


AaronErhardt:
Thanks, that a very good idea. Can you send a PR for "next" as well, even if it might be broken still?

euclio:
We probably want to just cherry-pick the commit? Will keep the history cleaner.

PR #156, opened 2022-04-03
Improvements to extension traits

cofee-on-the-desk:
- Add `ContainerChild` trait to define the type of container children. Previously this was inside `RelmRemovableExt`, which was fine but not all widgets that store children can actually remove them, for example `adw::TabView`. A few traits rely on this one to get the correct type of children.
- Add `RelmIterChildrenExt` impl for `adw::TabView`.
- Add `RelmRemoveAllExt` impl for T: `RelmSetChildExt`.
- Fix `RelmIterChildrenExt` impl for `gtk::Grid` (now relies on the widget position inside the grid to decide the right ordering).
- Move `RelmIterChildrenExt` to a separate module.
- Minor changes in tests

### DIFF
--- a/relm4/src/extensions/iter_children.rs
+++ b/relm4/src/extensions/iter_children.rs
@@ -1,0 +1,143 @@
+use super::ContainerChild;
+use crate::gtk;
+use gtk::prelude::{Cast, GridExt, IsA, WidgetExt};
+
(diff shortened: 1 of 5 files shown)


AaronErhardt (approved):
Good job. Looks very good :+1:

PR #157, opened 2022-04-03
Factory improvements

cofee-on-the-desk:
- Add `FactoryView` impl for `gtk::FlowBox`.
- Add tests for different widgets that impl `FactoryView`.
- Fix bugs related to the usage of `gtk::ListBoxRow` as the child inside `gtk::ListBox`.
- Fix bugs related to child ordering.
- Fix bugs related to invalid indexes.
- Minor changes in docs and examples.


AaronErhardt (approved):
Outstanding work! Thanks for all these improvements and fixes :slightly_smiling_face:

PR #158, opened 2022-04-03
fix layout of sudoku example

tronta:
not sure if it still makes sense to improve it in main, or if it makes more sense to port it to the new way…


AaronErhardt:
Let's merge it anyway, but of course porting it to the next branch would be great in the long term.
Thanks :)

PR #167, opened 2022-04-25
Add `sender()` methods to `RelmWorker` and `AsyncRelmWorker`

ids1024:
Presumably these types should have such a method, since `RelmComponent` does.


AaronErhardt:
Thanks!

PR #169, opened 2022-05-02
Implement `Debug` for all public types

euclio:
The overall goal of this PR is to make it possible to derive `Debug` for `Component` impls that have a `Controller` in them. This is accomplished by implementing `Debug` for all types in the library, including the complex types that were explicitly left unimplemented. This also involved relaxing the `Debug` bounds on `Sender` and `Receiver`, which weren't actually required.

Additionally, `Debug` is derived for the `Widgets` struct produced by the macro to make the new bounds easier to satisfy.

### DIFF
--- a/relm4/src/factory/mod.rs
+++ b/relm4/src/factory/mod.rs
@@ -22,9 +22,9 @@ use std::collections::hash_map::DefaultHasher;
 use std::collections::VecDeque;
 use std::hash::{Hash, Hasher};
 
-#[allow(missing_debug_implementations)]
 /// A container similar to [`VecDeque`] that can be used to store
 /// data associated with components that implement [`FactoryComponent`].
+#[derive(Debug)]
 pub struct FactoryVecDeque<Widget, C, ParentMsg>
 where
     Widget: FactoryView,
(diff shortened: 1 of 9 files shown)


relm4/src/factory/mod.rs
```diff
@@ -41,11 +41,14 @@ where
     uid_counter: u16,
 }
 
+#[allow(dead_code)] // As of 1.60 dead code analysis erroneously reports the Debug impl as dead
+#[derive(Debug)]
```
AaronErhardt, line 45: Can we investigate further why this is marked as dead code?

relm4/Cargo.toml
```diff
@@ -27,6 +27,7 @@ adw = { git = "https://gitlab.gnome.org/World/Rust/libadwaita-rs", optional = tr
 
 async-broadcast = "0.3.4"
 async-oneshot = "0.5.0"
+derivative = "2.2.0"
```
AaronErhardt, line 30: I see why it's useful, but I don't want to pull in a dependency that has more lines of code than the relm4 crate itself if implementing the traits manually is an option.

AaronErhardt (changes requested):
Thanks for the PR!
The code looks good, I'm just not sure whether `derivative` is necessary.

relm4/Cargo.toml
euclio: Fair, I'll post a new revision with manual impls.

relm4/src/factory/mod.rs
euclio: Absolutely, was planning to create an upstream bug report but didn't want to block this PR on that.

AaronErhardt (approved):
Thanks! Let's merge this regardless of the upstream error, like you suggested.

PR #184, opened 2022-05-15
only use `doc(cfg)` if `dox` cfg is present

### LINKED ISSUE #171: Unstable doc features should not use `cfg(doc)`
Gating unstable doc features like `doc_cfg` behind `cfg(doc())` is not a good solution, because `cfg(doc)` is always set by `rustdoc`, even on stable. This causes errors when running `cargo doc` on stable.

We should gate on an explicit cfg like `docsrs` and then only supply that `cfg` when building docs for the website, so that users can build the documentation locally without unstable features. [Here's what tokio does.](https://github.com/tokio-rs/tokio/blob/master/tokio/Cargo.toml#L148-L151)

euclio:
Fixes #171, which allows building the documentation with a stable compiler.

This will require changing the documentation build process to pass `--cfg dox` as a rustdoc argument.

### DIFF
--- a/relm4/src/factory/widgets/mod.rs
+++ b/relm4/src/factory/widgets/mod.rs
@@ -1,11 +1,11 @@
 mod gtk;
 
 #[cfg(feature = "libadwaita")]
-#[cfg_attr(doc, doc(cfg(feature = "libadwaita")))]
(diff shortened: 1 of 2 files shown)


AaronErhardt:
Thanks!

PR #196, opened 2022-05-29
Macro polish

AaronErhardt:
Stability improvements and syntax enhancements for 0.5


relm4-macros/src/factory/mod.rs
```diff
@@ -164,7 +164,8 @@ pub(crate) fn generate_tokens(
                 } = widgets;
 
                 // Wrap pre_view and post_view code to prevent early returns from skipping other view code.
-                (|| { #pre_view })();
+                #[warn(unreachable_code)]
+                #pre_view
```
euclio, line 168: Is this intentional?

PR #203, opened 2022-06-23
macros: Make view_output a function

AaronErhardt:
This is only to fix an incorrect rust-analyzer warning when using `view_output!()`.


AaronErhardt:
Not needed because this was fixed in newer versions of rust-analyzer

PR #204, opened 2022-06-28
tweak `WorkerController` and port non-blocking-async example

euclio:
This PR changes `WorkerController` and `WorkerHandle` to use the `Worker` impl instead of the input and output types directly, which reads a bit nicer (`WorkerController<MyWorker>` instead of `WorkerController<MyWorkerMsg>`.

This commit also ports over the non-blocking-async example.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -38,3 +38,7 @@ path = "progress.rs"
 [[example]]
 name = "macro_reference"
 path = "macro_reference.rs"
+
+[[example]]
+name = "non_blocking_async"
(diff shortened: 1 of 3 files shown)


AaronErhardt (changes requested):
Looks great and thanks for cleaning up the worker API.

I only have a few ideas to improve the example:
+ Window instead of ApplicationWindow
+ You can remove `append = &` (I think `set_child` can be removed as well) thanks to the RelmContainerExt trait

euclio:
Updated.

AaronErhardt (approved):
Thanks!

PR #206, opened 2022-07-03
spawn worker updates instead of awaiting them

euclio:
The current behavior will cause all updates to be serialized in the order they are received, instead of interleaving them. The change in behavior can be seen in the non-blocking async example by clicking the button three times.


AaronErhardt (approved):
Thanks!

PR #208, opened 2022-07-06
emit tracing logs on received inputs

### LINKED ISSUE #198: Log important events via tracing
In my components, I find it helpful to add a log statement to the beginning of the `update` implementation so that I can see what the received message was. It would be great if Relm4 could provide this logging for me.

I put together a quick POC using [tracing](https://github.com/tokio-rs/tracing) and it works really well with Relm4's event-driven nature! I created a new span in `update_with_view` so that each handled event is cleanly separated. Here's some sample log output from my application with the new logging:

```
update_with_view input=NewSelection(File("/home/euclio/repos/fm/Cargo.toml")), component="fm::AppModel"
 INFO received message: NewSelection(File("/home/euclio/repos/fm/Cargo.toml"))
 INFO new selection: "/home/euclio/repos/fm/Cargo.toml", last dir: "/home/euclio/repos/fm", diff: "Cargo.toml"

update_with_view input=NewSelection("/home/euclio/repos/fm/Cargo.toml"), component="fm::file_preview::FilePreviewModel"
 INFO received message: NewSelection("/home/euclio/repos/fm/Cargo.toml")
 INFO identified file as application/toml

update_with_view input=SelectionChanged("/home/euclio/pic"), component="fm::places_sidebar::PlacesSidebarModel"

update_with_view input=NewRoot("/home/euclio/pic"), component="fm::AppModel"
 INFO received message: NewRoot("/home/euclio/pic")
 INFO new root: "/home/euclio/pic"

update_with_view input=Hide, component="fm::file_preview::FilePreviewModel"
 INFO received message: Hide
```

Note that tracing logs are also compatible with the `log` crate, so we wouldn't be forcing downstream consumers to pick a logging implementation.

I'd love to explore this further, but before I do I had the following questions:

- Are we OK adding a `Debug` bound to `Input` and `Output` on `Component`? That's needed to be able to print arbitrary messages.
- Is there a better identifier to use for the component itself? `Debug` output is likely too verbose. My POC uses the `type_name`, but this is only helpful if the component is unique.

euclio:
Fixes #198.

Only logs `update_with_view` and `update_view_with_cmd`, expecting more events to be added in the future. Most importantly, this adds `Debug` bounds to `Input` and `Output` so that the breaking changes are in early.

### DIFF
--- a/relm4-macros/src/lib.rs
+++ b/relm4-macros/src/lib.rs
@@ -42,6 +42,7 @@ fn gtk_import() -> std::rc::Rc<syn::Path> {
 ///     counter: u8,
 /// }
 ///
+/// #[derive(Debug)]
 /// enum AppMsg {
 ///     Increment,
 ///     Decrement,
--- a/relm4-macros/tests/simple.rs
+++ b/relm4-macros/tests/simple.rs
@@ -6,6 +6,7 @@ struct AppModel {
     counter: u8,
 }
 
+#[derive(Debug)]
 enum AppMsg {
     Increment,
     Decrement,
--- a/relm4/Cargo.toml
+++ b/relm4/Cargo.toml
@@ -40,6 +40,7 @@ tokio = { version = "1.15", features = ["rt", "rt-multi-thread"] }
 
 #relm4-macros = { version = "0.4.1", optional = true }
 relm4-macros = { path = "../relm4-macros", optional = true }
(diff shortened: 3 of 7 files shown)


AaronErhardt (approved):
Looks good to me! I only wonder whether we should make this feature optional for those that don't want to implement `Debug`, but adding that later wouldn't be a breaking change so we can wait and see whether this will be requested.

PR #216, opened 2022-07-25
Factory guard update

cofee-on-the-desk:
- Mutable methods are now accessible only through `FactoryVecDequeGuard`. All changes to the factory are rendered when the guard is dropped.
- Simplify `guard` method. Now returns a guard directly instead of wrapping it inside an `Option`. There was no need to check if the guard was already created because that is handled compile-time by rust borrowing rules.
- `apply_external_changes` is now also called automatically when `FactoryVecDequeGuard` is created, so there should be no way for it to cause undefined behaviour. I also fixed a bug in which it the method won't recognize in some tabs if `adw::TabView` were closed.
- Adjust examples
- Add basic documentation

I believe this approach is a lot better than the current one. My only concern is that I had to explicitly drop the guard in `tab_game` example, which was a little unintuitive, I wonder how could we fix that.

### DIFF
--- a/examples/libadwaita/tab_game.rs
+++ b/examples/libadwaita/tab_game.rs
@@ -254,17 +254,21 @@ impl Component for AppModel {
             counters: FactoryVecDeque::new(widgets.tabs.clone(), &sender.input),
             start_index: None,
         };
+
+        let mut counters_guard = model.counters.guard();
         for i in 0..3 {
-            model.counters.push_back(i);
+            counters_guard.push_back(i);
         }
-        model.counters.render_changes();
+
+        // Explicitly drop the guard,
+        // so that 'model' is no longer borrowed
+        // and can be moved inside ComponentParts
+        drop(counters_guard);
 
         ComponentParts { model, widgets }
     }
 
     fn update(&mut self, msg: Self::Input, sender: &ComponentSender<Self>) {
-        self.counters.apply_external_updates();
-
         match msg {
             AppMsg::StartGame(index) => {
                 self.start_index = Some(index);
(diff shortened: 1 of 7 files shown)


examples/libadwaita/tab_game.rs
```diff
@@ -254,17 +254,21 @@ impl Component for AppModel {
...
         for i in 0..3 {
-            model.counters.push_back(i);
+            counters_guard.push_back(i);
         }
-        model.counters.render_changes();
+
+        // Explicitly drop the guard,
+        // so that 'model' is no longer borrowed
+        // and can be moved inside ComponentParts
+        drop(counters_guard);
```
AaronErhardt, line 266: The drop here is indeed not the nicest solution. It should work with scopes too. Can you try wrapping the "guarded" part into brackets {} ?

AaronErhardt (approved):
Well done! This turned out really well IMO :)

There's only this thing with the `drop` in tab_game but that can be implemented in another PR,

examples/libadwaita/tab_game.rs
cofee-on-the-desk: Makes sense, but I think with scopes it's still confusing.

AaronErhardt: It's not an uncommon pattern to do something like this in Rust. I know it's not the prettiest thing, but IMO still better than dropping manually.

```Rust
let mutex = Mutex::new();
// Critical section
{
    let guard = mutex.lock();
    // Do critical stuff
} // Make sure the guard is released
// Do non critical stuff
```

PR #217, opened 2022-07-26
core: Add method to drop factory guard

AaronErhardt:
It's not perfect, but maybe a little bit better. What do you think?


cofee-on-the-desk (approved):
Yeah, it's better. Maybe I would rephrase the documentation to make it more readable. 

```
    /// Drops the guard and renders all changes.
    ///
    /// Use this to transfer full ownership back to the [`FactoryVecDeque`].
```

PR #218, opened 2022-07-27
Make `FactoryComponent` API closer to `Component`

cofee-on-the-desk:


relm4-macros/src/factory/mod.rs
```diff
@@ -135,19 +135,23 @@ pub(crate) fn generate_tokens(
...
+        let mut args = match &last_segment.arguments {
+            syn::PathArguments::AngleBracketed(args) => Some(args.args.clone()),
+            _ => None,
         }
-        _ => None,
-    }
-    .unwrap_or_default();
+        .unwrap_or_default()
+        .into_iter();
+        (args.next().unwrap(), args.next().unwrap())
```
AaronErhardt, line 146: Unwraps are probably a bad idea at this point, because macros are supposed to handle errors properly and you never know how many arguments are actually there. Maybe there is a good default for those values (the unit type maybe`) or we can avoid generics at this point entirely.

AaronErhardt (commented):
Good job overall! I'm surprised how well the macro changes are implemented given that it's not easy to understand the code without prior experience with proc-macros. :+1:

relm4-macros/src/factory/mod.rs
cofee-on-the-desk: I'm not sure in which kind of situation this would panic, but I understand your concern. For now, let's use `()` as the default.

AaronErhardt (approved):
Thanks!

PR #219, opened 2022-07-28
Do not use generics in `FactoryComponent`

cofee-on-the-desk:
`FactoryComponent` uses generics to define `ParentWidget` and `ParentMsg`. This way, a single type can have multiple component implementations for different container types. But it comes with a tradeoff: generic arguments must be specified for most factory types. For example, `FactoryVecDeque<gtk::Box, Counter, AppMsg>` or `FactoryComponentSender<gtk::Box, AppMsg, Self>`.

This PR replaces generics with associated types, removing the need to specify generic arguments: `FactoryVecDeque<Counter>`, `FactoryComponentSender<Self>`.

It would still be possible to provide multiple implementations, if the type that implements `FactoryComponent` is itself generic.


AaronErhardt (approved):
Looks good, thanks!

Issue #100, opened 2022-02-02
Port examples to the "next" branch


AaronErhardt:
The **_bold and italic_** entries are most important, the rest could be replaced or added later. The libadwaita examples (not listed here) should be ported as some point, too.

- [ ] actions
- [x] alert
- [x] **_components_**
- [ ] components_old
- [x] **_drawing_**
- [ ] entry
- [ ] entry_tracker
- [x] **_factory_**
- [ ] ~~factory_advanced~~
- [ ] factory_advanced_manual
- [ ] factory_manual
- [ ] future
- [x] **_grid_factory_**
- [ ] list
- [x] **_macro_reference_**
- [x] **_menu_**
- [ ] micro_components
- [ ] micro_components_manual
- [x] non_blocking_async
- [ ] non_blocking_async_manual
- [x] open_button
- [ ] popover
- [x] save_dialog
- [x] **_simple_**
- [x] **_simple_manual_**
- [ ] stack
- [ ] stack_factory
- [ ] stack_switcher
- [ ] stateful_msg_handler
- [x] to_do
- [x] tokio
- [x] **_tracker_**

PR #221, opened 2022-07-28
Port tracker and to_do examples to 0.5

euclio:
cc #100 

This PR also adds a delete button to the to_do example, which helps demonstrate internal communication in the factory component versus communication with the parent.


AaronErhardt (changes requested):
Well done! I only would like to get rid of the `append`s and `set_child`s because the new syntax without them looks much cleaner IMO.

euclio:
Fixed.

AaronErhardt (approved):
Awesome, thanks!

PR #224, opened 2022-07-30
Pass `ComponentSender` by value

cofee-on-the-desk:


AaronErhardt (approved):
I ran some update function stress test benchmarks on this branch and apparently there's no measurable difference between cloning and passing references.

Thanks!

PR #226, opened 2022-07-31
Port drawing example to 0.5

euclio:
cc #100

### DIFF
--- a/examples/drawing.rs
+++ b/examples/drawing.rs
@@ -1,0 +1,197 @@
+use std::time::Duration;
+
+use gtk::cairo::{Context, Operator};
+use gtk::prelude::*;
+use relm4::drawing::DrawHandler;
+use relm4::{gtk, Component, ComponentParts, ComponentSender, RelmApp, WidgetPlus};
+
+#[derive(Debug)]
+enum AppMsg {
+    AddPoint((f64, f64)),
+    Reset,
+    Resize((i32, i32)),
+}
+
+#[derive(Debug)]
+struct UpdatePointsMsg;
+
+struct AppModel {
+    width: f64,
+    height: f64,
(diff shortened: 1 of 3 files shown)


examples/drawing.rs
```diff
@@ -0,0 +1,200 @@
...
+        let widgets = view_output!();
+
+        model.handler.init(&widgets.area);
+
+        sender.command(|out, shutdown| {
+            shutdown
+                .register(async move {
+                    loop {
+                        tokio::time::sleep(Duration::from_millis(20)).await;
+                        out.send(AppMsg::UpdatePoints);
```
AaronErhardt, line 136: You're normally not supposed to do so, but you could move a clone of the sender into the closure to send messages directly to input. Or even better, remove the AppMsg::UpdatePoints variant and update the points only from `update_cmd`.

AaronErhardt (changes requested):
Looks good, just one minor thing to improve.

examples/drawing.rs
euclio: > You're normally not supposed to do so, but you could move a clone of the sender into the closure to send messages directly to input.

I'm a bit wary of doing something you're "not supposed to do" in exemplary code. Maybe it's fine with a comment?

> remove the AppMsg::UpdatePoints variant and update the points only from `update_cmd`.

I tried this, but then a view update isn't triggered.

AaronErhardt: > I tried this, but then a view update isn't triggered.

Actually, `update_cmd` should trigger a view update.

euclio: You're right, I was missing a call to `draw`. Fixed, and simplified the reset logic a bit, too.

AaronErhardt (approved):
Thanks!

PR #230, opened 2022-08-02
Implement Factory for adw Carousel

vlinkz:
More libadwaita :)

### DIFF
--- a/relm4/src/factory/widgets/libadwaita.rs
+++ b/relm4/src/factory/widgets/libadwaita.rs
@@ -104,3 +104,58 @@ impl FactoryView for adw::ExpanderRow {
         self.factory_append(widget, position);
     }
 }
+
+impl FactoryView for adw::Carousel {
+    type Children = gtk::Widget;
+    type ReturnedWidget = gtk::Widget;
+    type Position = ();
+
+    fn factory_remove(&self, widget: &Self::ReturnedWidget) {
+        self.remove(widget);
+    }
+
+    fn factory_append(
+        &self,
+        widget: impl AsRef<Self::Children>,
+        _: &Self::Position,
+    ) -> Self::ReturnedWidget {
+        self.append(widget.as_ref());
+        widget.as_ref().clone()
+    }
+
+    fn factory_prepend(
+        &self,
+        widget: impl AsRef<Self::Children>,
+        _position: &(),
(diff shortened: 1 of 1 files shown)


relm4/src/factory/widgets/libadwaita.rs
```diff
@@ -104,3 +104,51 @@ impl FactoryView for adw::ExpanderRow {
...
+    ) -> Self::ReturnedWidget {
+        self.factory_append(widget.as_ref(), position);
+        widget.as_ref().clone()
+    }
+
+    fn returned_widget_to_child(returned_widget: &Self::ReturnedWidget) -> Self::Children {
+        returned_widget.clone()
+    }
+
+    fn factory_move_after(&self, _widget: &Self::ReturnedWidget, _other: &Self::ReturnedWidget) {}
```
AaronErhardt, line 149: Can we implement this method somehow? Maybe we can find the position of the other widget and then use `reorder` to move it. Btw. I'm just on my phone right now but I'll have a closer look at this PR after work.

vlinkz: It could be possible, but the operation would take O(n) time. Currently, the only suitable way I see to index widgets would be to iterate through all of them using [adw::Carousel::nth_page](https://world.pages.gitlab.gnome.org/Rust/libadwaita-rs/stable/latest/docs/libadwaita/struct.Carousel.html#method.nth_page), check equality, and then reorder to the iterator once a match is found

AaronErhardt: I wouldn't mind. We're not writing some high performance code here and if we don't implement it, the factory won't work as expected. The other move after implementations will probably be O(n) too.

AaronErhardt (changes requested):
I'm in favor of implementing the factory correctly for several reasons:

1. Figuring out the position of a widget (iterating over the children) from a Carousel is o(n) but still cheap.
2. I'm pretty sure GTK's reorder operations that are used in other `FactoryView` impls are O(n) too.
3. It's nice to have the widgets reordered when you reorder the data.

If we don't do that, it would be better remove the `factory_move_start` impl too because otherwise it's only half-functional.

AaronErhardt (approved):
Thanks!

PR #233, opened 2022-08-03
Miscellaneous improvements

AaronErhardt:
+ Fix all clippy warnings due to libadwaita feature flags
+ Remove `Sync` requirement for command closures
+ Add convenience method for spawning futures that will return one `CommandOutput`

Arguably, the last point is debatable and I'm not entirely sure whether this method could be implemented or named better.


relm4/src/component/sender.rs
```diff
@@ -45,6 +47,21 @@ impl<Input, Output, CommandOutput: Send + 'static>
         });
     }
 
+    /// Spawn a future that will be dropped as soon as the component is shut down.
+    ///
+    /// Essentially, this is a simpler version of [`command()`].
+    pub fn simple_command<Fut>(&self, future: Fut)
```
euclio, line 53: Name bikeshed: `oneshot_command`, `command_fut`

mmstick:
What if we make the API a bit more opinionated to simplify it? Perhaps commands should be dropped by default anyway, and if you want something to outlive the component, you should use `relm4::spawn`.

AaronErhardt:
I like the idea with making the API more opinionated. That's the idea behind the simplified command version anyway. Yet we have one problem: As long as we don't have async closures, we cannot pass the sender to the future so the user has to clone and move the sender by themselves. So for more complex examples (for example where you send multiple messages) it is nice to have the closure to produce the future, I guess.

relm4/src/component/sender.rs
```diff
@@ -45,12 +47,32 @@ impl<Input, Output, CommandOutput: Send + 'static>
         });
     }
 
+    /// Spawns a future that will be dropped as soon as the component is shut down.
+    ///
+    /// Essentially, this is a simpler version of [`Self::command()`].
+    ///
+    /// ```rust
+    ///
```
euclio, line 55: Missing example here

euclio (approved):

PR #242, opened 2022-08-13
core: Add message broker

### LINKED ISSUE #213: Communicate between components on different levels
### Discussed in https://github.com/Relm4/Relm4/discussions/212

<div type='discussions-op-text'>

<sup>Originally posted by **Paveloom** July 18, 2022</sup>
So, imagine you have the following graph of components:

```mermaid
flowchart LR
Window --> A --> B --> C
Window --> Dialog
```

The dialog is set to be modal and transient for the parent window, so it can't be moved. However, its sole purpose is to communicate with the `C` component, so it would naturally belong as a component of the `C` component. There are basically two diametrical problems here:

1) If the graph is left as is, one should construct a channel between the two components by creating messages in *each* of the components in between: this pollutes the code of the components and leads to quite a lot of duplication.
2) If the `Dialog` component is moved to be a component of the `C` component, there is no way to make it transient for the parent window. Well, one could copy the parent window into each component (similar to the [`save_dialog`](https://github.com/Relm4/Relm4/blob/8985b76ae0ac612b63c0830fb8c6ec85327c3af2/relm4-examples/examples/save_dialog.rs) example, but saving the window into an additional field of *each* component in between), but that's just a waste of memory (also, pollutes the model of each component).

Are there nicer ways to handle either of these problems?</div>

AaronErhardt:
#### Summary

Fixes #213 

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

@Paveloom what do you think? Do you want to review this PR?

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -73,4 +73,4 @@ jobs:
       - uses: codespell-project/actions-codespell@master
         with:
           check_filenames: true
-          ignore_words_list: "crate"
+          ignore_words_list: crate,statics
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
++ core: Add MessageBroker type to allow communication between components on different levels
 + core: Rename InitParams to Init in SimpleComponent and Worker too
 
 ## 0.5.0-beta.2 - 2022-8-12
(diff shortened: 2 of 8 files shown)


Paveloom:
LGTM. The only thing is that the example is not configured in `examples/Cargo.toml`.

I haven't upgraded my project to `v0.5` yet, but I will do it soon. Thank you very much for implementing this.

AaronErhardt:
Thanks! 
I accidentally modified `examples/Cargo.toml` in another PR already so it should work fine once merged.

PR #244, opened 2022-08-13
Refactor factory macro to use visitors

euclio:
#### Summary

Follow-up to #207. This PR refactors the factory code to use visitors like the component macro, making it easier for the macros to share code. There are still plenty more opportunities for re-use, but this PR lays the groundwork and demonstrates how the pre- and post- view parsing code can be reused between the two components. Eventually I'd like to have these visitors do the token generation too so that code can be shared as well.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


relm4-macros/src/visitors.rs
```diff
@@ -98,26 +110,128 @@ impl VisitMut for ComponentVisitor {
...
-                    let existing = self.pre_view.replace(func.clone());
-                    if existing.is_some() {
-                        self.errors
-                            .push(syn::Error::new_spanned(func, "duplicate pre_view function"));
+            }
+            _ => (),
+        }
+
+        if remove {
+            *item = syn::ImplItem::Verbatim(quote! {});
```
AaronErhardt, line 118: Since you've already defined a function for that, this should be `null_item()` too.

AaronErhardt (approved):
Looks good!

relm4-macros/src/visitors.rs
euclio: Fixed.

AaronErhardt (approved):
Thanks!

PR #254, opened 2022-08-20
core: Use late init for set_transient_for

AaronErhardt:
#### Summary

`set_transient_for` required a `Window` so far, which requires you to pass down the window from component to component which is quite tedious. The idea behind this PR is the following: You could use `WidgetExt::Root` to get the root window from any widget so you don't have to pass the window around. However, this requires the root widget of the component to be connected in the widget tree. The current initialization order prevents this because components run their `init` function often before their root widget was connected by their parent component. To solve this issue, we could run a late initiation pass once all widgets have been initialized and connected.

@Paveloom what are your thoughts?

Fixes an issue discussed in #245 

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


Paveloom:
I don't see it working. I'm trying

```rust
dbg!(root.root());
let dialog = Dialog::builder()
            .transient_for(root)
            .launch_with_broker((), &DIALOG_BROKER)
            .forward(sender.input_sender(), identity);
```

in the [example](https://github.com/Relm4/Relm4/discussions/245#discussioncomment-3396902). The debug call returns `false`, so the code in the new `transient_for` doesn't run.

relm4/src/component/builder.rs
```diff
@@ -66,15 +66,25 @@ where
...
     /// Set the component's root widget transient for a given window.
     ///
     /// If the root widget is a native dialog, such as [`gtk::FileChooserNative`],
     /// you should use [`transient_for_native`][ComponentBuilder::transient_for_native] instead.
     #[must_use]
-    pub fn transient_for(self, window: impl AsRef<gtk::Window>) -> Self {
-        self.root.as_ref().set_transient_for(Some(window.as_ref()));
+    pub fn transient_for(self, widget: impl AsRef<gtk::Widget>) -> Self {
+        let root = widget.as_ref().root();
+        if let Some(root) = root {
```
Paveloom, line 78: Also, what if this `root` isn't a window? In the [example](https://github.com/Relm4/Relm4/discussions/245#discussioncomment-3396902) it's actually a button.

relm4/src/component/builder.rs
```diff
@@ -66,15 +66,25 @@ where
...
+            let window = root.downcast::<gtk::Window>().unwrap();
```
Paveloom, line 79: Also, what if this root isn't a window? In the [example](https://github.com/Relm4/Relm4/discussions/245#discussioncomment-3396902) it's actually a button.

AaronErhardt: `WidgetExt::root` is not the root of the component, but the next root widget in the widget tree (such as Window). But I replaced this with `toplevel_window` which is the better function probably.

AaronErhardt:
Yes, there were some major logical flaws in my code but they should be fixed and I added your example too to see whether it works and it actually does now!

Paveloom:
Aight, I can confirm it works. Thank you.

Gotta update the docstring, though. Also, would this code apply to [native dialogs](https://relm4.org/docs/next/relm4/component/struct.ComponentBuilder.html#method.transient_for_native), too?

AaronErhardt:
Native dialogs should work now and the docs are also updated.

Paveloom:
LGTM. Thank you.

PR #259, opened 2022-08-21
Return a clone of the `DynamicIndex` after inserting into a factory

### LINKED ISSUE #257: Return a clone of the `DynamicIndex` after inserting a new item into a factory
I just hit a use case where I need to store dynamic indices of new items in the models of other items in the factory, but there is no way to obtain a `DynamicIndex` right now, except through a connected event. I would like to suggest to return a clone of the dynamic index in all `insert` calls. For example, a clone of the `dyn_index` here:

https://github.com/Relm4/Relm4/blob/ae761b14064ead777137637db51a60158dbf1433/relm4/src/factory/collections/vec_deque.rs#L189-L220

Paveloom:
#### Summary

Gonna need it myself anyway. Fixes #257.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [ ] cargo test*
- [x] updated CHANGES.md

\* On an unrelated note, compiling `libadwaita-examples` results in a memory leak and a crash on my system...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 + core: Add getter for global application for easier quitting of applications
 + core: Add MessageBroker type to allow communication between components on different levels
 + core: Rename InitParams to Init in SimpleComponent and Worker too
++ core: Return a clone of the `DynamicIndex` after inserting into a factory
 + macros: Fix returned widgets assignment in the view macro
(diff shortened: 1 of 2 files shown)


AaronErhardt (approved):
Thanks!

PR #260, opened 2022-08-24
always use `gtk::Application` in public APIs

### LINKED ISSUE #255: Decouple `libadwaita` feature from main application type
Some applicaitons may want to use Relm implementations (such as factories) on libadwaita widgets without using the libadwaita Application type. It would be nice to decouple the two, so that a savvy user could use `set_application` to use `gtk::Application` instead.

euclio:
#### Summary

This PR contains a few improvements to how we handle GTK application types. Chiefly, it replaces the `Application` type alias with `IsA<gtk::Application>`, because `adw::Application` is a subclass of `gtk::Application`. This, along with a change to not manually initialize GTK or libadwaita, allows users to enable the libadwaita feature without being forced into using `adw::Application` or libadwaita styles. Lastly, this PR makes the `app` field on `RelmApp` private, because it can be accessed through `relm4::main_application`.

Fixes #255.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/relm4/src/lib.rs
+++ b/relm4/src/lib.rs
@@ -86,14 +86,6 @@ pub use panel;
 pub use once_cell;
 pub use tokio;
 
-/// Initialize GTK and (optionally) libadwaita
-fn init() {
-    gtk::init().unwrap();
-
-    #[cfg(feature = "libadwaita")]
-    adw::init();
-}
-
 thread_local! {
     static MAIN_APPLICATION: Cell<Option<gtk::Application>> = Cell::default();
 }
--- a/relm4/src/lib.rs
+++ b/relm4/src/lib.rs
@@ -52,6 +52,7 @@ pub use app::RelmApp;
 pub use tokio::task::JoinHandle;
 pub use util::{WidgetPlus, WidgetRef};
 
+use gtk::prelude::*;
 use once_cell::sync::OnceCell;
 use std::cell::Cell;
 use std::future::Future;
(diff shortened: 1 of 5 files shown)


relm4/src/lib.rs
```diff
@@ -51,6 +51,7 @@ pub use app::RelmApp;
 pub use tokio::task::JoinHandle;
 pub use util::{WidgetPlus, WidgetRef};
 
+use gtk::prelude::*;
 use once_cell::sync::OnceCell;
```
AaronErhardt, line 55: Are there so many imports that it's worth to use the wildcard here?

AaronErhardt (changes requested):
Looks good to me. This was definitely a good idea, changing type definitions depending on feature flags was not ideal.

There's only one import that should be cleaned up.

Edit: sorry that I messed up the CI, I was already too tired I guess and should have done this tomorrow...

AaronErhardt:
Also, don't forget to pull the merge commit that sorts out the conflicts.

relm4/src/lib.rs
euclio: It could go either way, but for a prelude import I usually prefer to just glob import it so that all the convenient types are already in scope.

AaronErhardt: It*s surely more convenient but IMO the code is easier to read when you know which types are in scope and from which import they come from.

AaronErhardt (approved):
Thanks!

PR #261, opened 2022-08-25
Speedup CI by linking with mold

AaronErhardt:


AaronErhardt:
This breaks because `jq` is not installed on the system. I'm trying to get mold setup in the docker image instead.

PR #264, opened 2022-08-29
examples: Add optional assign to macro_reference

### LINKED ISSUE #263: Add example of optional property syntax to macro_reference example
The [optional property syntax](https://relm4.org/book/stable/widget_macro_reference.html#assign-properties) in the `view!` macro (`optional_property?: some_value`) appears to still be supported, but we don't have an example of it anywhere in the repository. Perhaps it's less useful now that gtk-rs' methods tend to not take `Option`s anymore?

We should either add an example of it to the `macro_reference` example, or remove support for it.

AaronErhardt:
#### Summary

Fixes #263 

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -36,17 +36,14 @@ jobs:
       - name: "clippy all features"
         run: cargo clippy --all-features -- --deny warnings
 
-      # nightly
-      - name: "test nightly"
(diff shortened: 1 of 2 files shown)


euclio (approved):
Thank you!

PR #270, opened 2022-09-11
port alert example to 0.5

euclio:
#### Summary

Port the `alert` example to 0.5 syntax. cc #100.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/examples/relm4-components/Cargo.toml
+++ b/examples/relm4-components/Cargo.toml
@@ -11,6 +11,10 @@ relm4 = { path = "../../relm4", features = ["macros"] }
 relm4-components = { path = "../../relm4-components" }
 
 [[example]]
+name = "alert"
+path = "alert.rs"
+
+[[example]]
 name = "file_dialogs"
 path = "file_dialogs.rs"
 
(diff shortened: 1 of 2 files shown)


examples/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2021"
 
 [dev-dependencies]
 relm4 = { path = "../relm4", features = ["macros"] }
+relm4-components = { path = "../relm4-components" }
```
AaronErhardt, line 9: I think we should rather place it here, right? https://github.com/Relm4/Relm4/tree/main/examples/relm4-components

AaronErhardt (changes requested):
LGTM. It just makes sense to use the relm4-components folder for related examples since we already have it.

examples/Cargo.toml
euclio: Yes, fixed.

AaronErhardt (approved):
Thanks!

PR #271, opened 2022-09-11
add a prelude module

euclio:
#### Summary

This PR adds a prelude module that contains the traits and types that are needed to implement a component. This helps reduce the import boilerplate required to implement a new component. Two examples are modified to demonstrate the benefit.

I purposely tried to keep it pretty conservative. Right now it only contains traits, the types required to implement the methods on those traits, and the gtk crates. It might be helpful to include `RelmApp` or `FactoryVecDeque`, but those are left out for now.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/examples/to_do.rs
+++ b/examples/to_do.rs
@@ -1,6 +1,6 @@
 use gtk::prelude::*;
-use relm4::factory::{DynamicIndex, FactoryComponent, FactoryComponentSender, FactoryVecDeque};
-use relm4::{gtk, ComponentParts, ComponentSender, RelmApp, SimpleComponent, WidgetPlus};
+use relm4::factory::FactoryVecDeque;
+use relm4::prelude::*;
 
 #[derive(Debug)]
 struct Task {
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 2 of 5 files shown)


examples/to_do.rs
```diff
@@ -1,6 +1,7 @@
 use gtk::prelude::*;
-use relm4::factory::{DynamicIndex, FactoryComponent, FactoryComponentSender, FactoryVecDeque};
-use relm4::{gtk, ComponentParts, ComponentSender, RelmApp, SimpleComponent, WidgetPlus};
+use relm4::factory::FactoryVecDeque;
+use relm4::prelude::*;
+use relm4::RelmApp;
```
AaronErhardt, line 4: I think we should probably add RelmApp to the prelude too. What do you think?

AaronErhardt (approved):
Looks good. Let me know when you think it's ready.

examples/to_do.rs
euclio: Sure, that one seems relatively low-risk since the type name is already "namespaced".

euclio: Added.

AaronErhardt (approved):
Thanks!

PR #272, opened 2022-09-11
reimplement senders as structs instead of type aliases

euclio:
#### Summary

This PR changes the `ComponentSender` and `FactoryComponentSender` types to be true structs instead of type aliases of `Arc<ComponentSenderInner>`. This has a number of benefits:

- The documentation for `ComponentSender` is more helpful.
- We can make `ComponentSenderInner` private and avoid exposing its internals. This will give us greater flexibility in the future to change the underlying representation. It would be nice to make `input` and `output` private in favor of the accessor methods, but they are left public for now to avoid making another large breaking change before 0.5.
- The error messages containing `ComponentSender` are nicer, because they contain the name of the actual struct instead of the alias (which is a very long type name).

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


relm4/src/sender.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// Emit an input to the component.
+    pub fn input(&self, message: C::Input) {
+        self.input.send(message)
+    }
+
+    /// Equivalent to `&self.input`.
+    #[must_use]
+    pub fn input_sender(&self) -> &Sender<C::Input> {
+        &self.input
+    }
```
AaronErhardt, line 75: I'd say we either make this field public xor have a getter method. I think it'd usually be better to hide the internal fields...

relm4/src/sender.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+impl<C: Component> Clone for ComponentSender<C> {
+    fn clone(&self) -> Self {
+        ComponentSender {
+            input: self.input.clone(),
+            output: self.output.clone(),
+            command: self.command.clone(),
+            shutdown: self.shutdown.clone(),
```
AaronErhardt, line 95: I think this isn't great from a performance perspective. Most sender types use atomics internally so it would be better to have something similar to the old `Arc<ComponentSenderInner>` to avoid multiple atomic operations (it's not a big deal actually but should be an easy optimization).

AaronErhardt (changes requested):
Overall I think this is a very good idea. Maybe we can keep `ComponentSenderInner` or something similar (as a hidden item this time) in order to slightly reduce code duplication and the amount of atomic operations.

relm4/src/sender.rs
euclio: Agreed. If we're fine making another large breaking change before 0.5, then I'd propose just removing `input` and `output` in favor of `input_sender` and `output_sender`, though this will likely involve updating almost every `Component` impl out there.

Current status:

| Retrieve input sender | Retrieve output sender | Send input | Send output |
| ----------------------------- | -------------------------------| ----------- | --------|
| `.input` or `.input_sender()` | `.output` or `.output_sender()` | `.input(msg)` | `.output(msg)` |

Proposed:

| Retrieve input sender | Retrieve output sender | Send input | Send output |
| ----------------------------- | -------------------------------| ----------- | --------|
| `.input_sender()` | `.output_sender()` | `.input(msg)` | `.output(msg)` |

Removing the `input` and `output` fields will also allow us to bring back `ComponentSenderInner` to share implementation between the two.

AaronErhardt: The proposed changes sound good to me. Also I don't think the senders are often used directly, but by sending messages with `.input()` for example. So I don't think it will affect a lot of existing 0.5 projects. Yet in any case it's not a very big breaking change either way.

euclio: Implemented!

relm4/src/sender.rs
euclio: Agreed, fixed by adding back `ComponentSenderInner`.

AaronErhardt (approved):
Thanks!

PR #278, opened 2022-09-14
core: Remove Debug requirement for FactoryComponent

### LINKED ISSUE #273: `FactoryComponent` should not require `Debug`
The `Debug` requirement on the `FactoryComponent` trait is only used for the `Debug` impl of `FactoryHandle`. It should be removed to give the user more flexibility.

AaronErhardt:
#### Summary

Fixes #273 

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/relm4/src/factory/handle.rs
+++ b/relm4/src/factory/handle.rs
@@ -22,7 +22,7 @@ pub(super) struct FactoryHandle<C: FactoryComponent> {
 impl<C: FactoryComponent> fmt::Debug for FactoryHandle<C> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("FactoryHandle")
-            .field("data", &self.data)
+            .field("data", &"<FactoryComponent>")
             .field("root_widget", &self.root_widget)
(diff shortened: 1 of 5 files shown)


relm4/src/factory/handle.rs
```diff
@@ -22,7 +22,6 @@ pub(super) struct FactoryHandle<C: FactoryComponent> {
 impl<C: FactoryComponent> fmt::Debug for FactoryHandle<C> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("FactoryHandle")
-            .field("data", &self.data)
```
euclio, line 25: We may want to leave this as `.field("data", &"<component>")` or similar instead.

euclio (approved):

relm4/src/factory/handle.rs
AaronErhardt: Done

euclio: LGTM!

PR #280, opened 2022-09-16
Add chain attr to properties

AaronErhardt:
#### Summary

Adds a `#[chain(method())]` attribute to work with types returned by property assignments. For example, `bind_property()` returns a builder that needs a call to `build()` in order to take effect.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
++ macros: Improve error reporting on invalid trait implementations
++ macros: Add chain attribute for properties
 + core: Impl extension traits and `FactoryView` for `adw::PreferencesGroup`
 + components: Add `SimpleComboBox` type as a more idiomatic wrapper around `gtk::ComboBoxText`
 + core: Append children for `gtk::Dialog` to its content area instead of using `set_child`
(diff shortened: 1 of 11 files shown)


CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
++ macros: Improve error reporting on invalid trait implementations
```
euclio, line 5: Can you add a UI test that demonstrates the improvement?

euclio (changes requested):
This is a cool idea. LGTM with a minor fix.

CHANGES.md
AaronErhardt: Old error message: 13fdd5fd2b8587e8febdca60e9ec7a5ab0a60820
New error message: 9b5f168508c58c6f4919544da55133a2f0d9375c

The new error reporting encourages rustc to further analyze the code. It is not better in every case, but I hope that it will help rust-analyzer to offer more completions in the future due to less code changed by the macro.

euclio (approved):

PR #282, opened 2022-09-21
macros: Don't generate dead code in widgets struct

AaronErhardt:
#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


euclio (commented):
Looks good. We could probably test this by writing an integration test with [`assert_fields`](https://docs.rs/static_assertions/latest/static_assertions/macro.assert_fields.html).

PR #285, opened 2022-09-24
core: Add convenience getters to Controller

AaronErhardt:
#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


relm4/src/component/controller.rs
```diff
@@ -22,6 +23,18 @@ pub trait ComponentController<C: Component> {
     /// Provides access to the state of a component.
     fn state(&self) -> &Rc<StateWatcher<C>>;
 
+    /// Returns a reference to the [`Component`].
+    fn get_model(&self) -> Ref<'_, C> {
```
euclio, line 27: I don't usually see the "get" prefix used in Rust. Any reasons these can't be `model()` and `widgets()`?

Or perhaps we should make the `Ref` more obvious with `model_ref()` and `widgets_ref()`?

AaronErhardt: Fixed

euclio (approved):

Issue #292, opened 2022-10-03
Builds on docs.rs are broken


pentamassiv:
### Open tasks

- [x] A PR for 0.5 (relm4 and relm4-components) https://github.com/Relm4/Relm4/pull/316
- [x] A backport to the stable branch for 0.4 https://github.com/Relm4/Relm4/pull/293
- [x] ~~A new release for 0.4 (I can take care of that)~~
- [ ] The stable docs generated by relm4-docs should show a deprecation warning (I think nightly docs are still kinda nice to have)
- [x] We should move some of the [pre-proccessed doc headers](https://relm4.org/docs/stable/relm4/) into the source code (especially the link collection is pretty useful I think). https://github.com/Relm4/Relm4/pull/338
- [ ] Update the links on the website and in the README (and everywhere else, probably some links in the book too)

### Why is this important?
Whenever a crate is published on crates.io, its documentation is automatically built and linked. Since you added a documentation attribute to the Cargo.toml files, the default link to docs.rs is replaced by that link. I am sure you had some reasons to do that but I am not a fan of it. I know that all gtk crates do that too, so maybe that was the idea. The downside of this is that the documentation is only available in the current version. If I use an older version, I can't have a look at its documentation (or it is not obvious how to do that). Luckily it is always possible to just go to `docs.rs/crate/CrateName` and view the documentation there. On docs.rs it is possible to look at all previous versions. This is why there is value in having working builds on docs.rs.

### What is the issue?
Unfortunately the build of the current version is broken. I had a similar problem with one of my crates and I believe the reason is that the dependencies are not available. This should be possible to be fixed by adding something along the lines of

```
[package.metadata.docs.rs]
features = ["dox"]

[features]
dox = ["gtk/dox", "adw/dox"]
```

to the `Cargo.toml` file. The same also applies to relm-components. relm-macros builds fine.
It is possible to test the build locally by following this guide:
https://github.com/rust-lang/docs.rs/blob/master/README.md#build-subcommand

### Explanation of the proposed solution
Here is my understanding of how this fixes the error:
```
[package.metadata.docs.rs]
features = ["dox"]
```
This tells docs.rs to activate the feature `dox`.
```
[features]
dox = ["gtk/dox", "adw/dox"]
```
If the feature `dox` is active, it activates it in the crates gtk4 and libadwaita too. This features allows building the docs without the dependencies.

pentamassiv:
[...]

PR #293, opened 2022-10-03
Fixed builds on docs.rs and bumped version

pentamassiv:
#### Summary
Backport to fix https://github.com/Relm4/Relm4/issues/292

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "relm4"
 authors = ["Aaron Erhardt <aaron.erhardt@t-online.de>"]
 edition = "2021"
-version = "0.4.4"
+version = "0.4.5"
 description = "An idiomatic GUI library inspired by Elm and based on gtk4-rs"
 repository = "https://github.com/Relm4/relm4"
 readme = "README.md"
@@ -19,6 +19,10 @@ members = [
     "relm4-examples/libadwaita",
     "relm4-macros",
 ]
+
+[package.metadata.docs.rs]
+features = ["dox"]
+cargo-args = ["-Zunstable-options", "-Zrustdoc-scrape-examples=examples"]
 
 [dependencies]
 adw = { version = "0.1", optional = true, package = "libadwaita" }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -37,11 +41,17 @@ tokio = { version = "1.15", optional = true, features = ["rt"] }
 
 [features]
 default = []
+dox = ["gtk/dox", "adw/dox"]
 libadwaita = ["adw"]
 libpanel = ["panel"]
```
AaronErhardt, line 46: I think we should add libpanel to the dox feature too if possible.

Cargo.toml
```diff
@@ -20,6 +20,10 @@ members = [
     "relm4-macros",
 ]
 
+[package.metadata.docs.rs]
+features = ["dox"]
```
AaronErhardt, line 24: I wonder whether it makes sense to use `all-features = true`. Otherwise, the code that is behind the "libadwaita" and "libpanel" feature flags won't be part of the documentation.

AaronErhardt (changes requested):

pentamassiv:
Building with the `dox` feature enabled requires a Nightly toolchain. Otherwise there will be this error:
```
error[E0554]: `#![feature]` may not be used on the stable release channel
  --> /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/glib-sys-0.15.10/src/lib.rs:12:30
   |
12 | #![cfg_attr(feature = "dox", feature(doc_cfg))]
   |                              ^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0554`.
error: could not compile `glib-sys` due to previous error
```
All other features continue to build fine with a stable toolchain. Should that be mentioned somewhere?

Cargo.toml
pentamassiv: Good idea

Cargo.toml
pentamassiv: I missed that. I didn't know the dependency was also gtk related.

AaronErhardt:
I think we should mention in the Cargo.toml file that the dox feature is just for docs.rs and requires nightly.

Also, can you disable the workflow for 1.56? This MSRV check isn't helpful anymore because some crates require higher Rust versions without raising the version requirement (which is ok, but annoying in this case). Maybe we need to add a Cargo.lock at some point...

pentamassiv:
Ah, now I understand. Done

AaronErhardt (approved):
Looks good to me! This should be all that's necessary to get out a new stable release, right?

pentamassiv:
Yes, I believe that's it

PR #297, opened 2022-10-04
all: Remove documentation field in favor of docs.rs

AaronErhardt:
I think this is still missing for getting docs.rs working @pentamassiv 

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ description = "An idiomatic GUI library inspired by Elm and based on gtk4-rs"
 repository = "https://github.com/Relm4/relm4"
 readme = "README.md"
 license = "Apache-2.0 OR MIT"
+documentation = "https://docs.rs/relm4/"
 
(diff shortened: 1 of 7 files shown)


pentamassiv:
I think it would be better to keep it and link to docs.rs. The reason for this is some odd behavior on crates.io. When you search for a crate and it does not have the documentation link set, in the search results there will no link to the docs. You will need to click on the crate and on that page, you can click on the documentation. It's just a minor nuisance though, so it's not that important.
You can see this behavior when you search for base64id (just a random crate I quickly found) on [crates.io](crates.io) for example.

This for example should work and then the documentation can be opened from the search results without going through the details page:
`documentation = "https://docs.rs/relm4/"`

AaronErhardt:
I've updated the links to point to docs.rs. But let's wait and see whether the docs actually work before merging this :sweat_smile:

PR #306, opened 2022-10-08
add `ComponentBuilder` to the prelude

euclio:
#### Summary

Adds `ComponentBuilder` to the prelude.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


AaronErhardt:
Which use-case do you have for `ComponentBuilder` in the prelude? It should be fine to call `MyComponent::builder()` to get the `ComponentBuilder` so you don't need to have it in scope.

euclio:
Whoops, you're right. Closing.

PR #308, opened 2022-10-09
improve macro error messages with missing `view!`

euclio:
#### Summary

This PR improves the macro error messages if the impl is missing a `view!` output invocation.

Previously, if you forgot to write the `view!` macro, the proc macro would confusingly say that the impl is missing a `init`/`init_widgets` function, even if you wrote one. This is because those functions would be removed as part of the component visitor, and only added back if we were able to parse the view widgets successfully. This is fixed by expanding `view_output!()` in-place, meaning that `init` and `init_widgets` are never removed from the impl, only mutated.

Secondly, this PR adds an error message if the macro was unable to find the `view!` macro, making the real issue clearer.

Lastly, this PR implements `Debug` for all macro visitors to assist debugging.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [ ] updated CHANGES.md


relm4-macros/src/visitors.rs
```diff
@@ -443,6 +425,81 @@ impl VisitMut for PreAndPostView<'_> {
...
+        {
+            if let syn::Expr::Macro(mac) = &**expr {
+                if mac.mac.path.is_ident("view_output") {
+                    expand = true;
+                }
+            }
+
+            if expand {
+                mem::swap(expr, &mut self.widgets_init);
+            }
```
AaronErhardt, line 487: What's the point of the `mem::swap` here? It looks to me like it is there to avoid uninitialized memory, but maybe you can add a comment here so it's not as confusing to read.

AaronErhardt (changes requested):
Looks very good!
There's just one line where a comment would be great and the merge conflicts that should be addressed.

relm4-macros/src/visitors.rs
euclio: Fixed.

AaronErhardt (approved):
Thanks!

PR #309, opened 2022-10-09
add responsive sidebar example using adw::Leaflet

euclio:
#### Summary

This PR creates a new example that uses `adw::Leaflet` for a responsive sidebar.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/examples/libadwaita/leaflet_sidebar.rs
+++ b/examples/libadwaita/leaflet_sidebar.rs
@@ -1,0 +1,135 @@
+/// Responsive sidebar layout inspired by the example code in the [libadwaita documentation].
+///
+/// Shrink the window small enough to see the sidebar and content pages become folded.
+///
+/// [libadwaita documentation]: https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/adaptive-layouts.html#leaflet
+
+use adw::prelude::*;
+use gtk::glib;
+use relm4::prelude::*;
+
+struct App {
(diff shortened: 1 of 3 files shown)


examples/libadwaita/leaflet_sidebar.rs
```diff
@@ -0,0 +1,135 @@
...
+                &widgets.content_header,
+                "show-start-title-buttons",
+            )
+            .flags(glib::BindingFlags::SYNC_CREATE)
+            .build();
+        widgets
+            .leaflet
+            .bind_property("folded", &widgets.back_button, "visible")
+            .flags(glib::BindingFlags::SYNC_CREATE)
+            .build();
```
AaronErhardt, line 122: Technically, this could be expressed within the `view` macro now, but probably it's better to keep it here to make it more readable.

AaronErhardt (changes requested):
LGTM. Can you add a changelog entry for this? So far, adding new examples was always mentioned.

examples/libadwaita/leaflet_sidebar.rs
euclio: Agreed. I had trouble getting `chain` to work with `flags`.

euclio:
Added changelog entry.

AaronErhardt (approved):
Thanks!

PR #311, opened 2022-10-09
core: Add ComponentStream type

AaronErhardt:
#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


AaronErhardt:
It was a good idea to add an example, there were several hidden problems with the first implementation. For example, the stream wasn't `Send` and therefore unusable in commands. The `CommandStream` type is now pretty reduced, but does exactly what it should.

While implementing the example, I also fixed a problem with `transient_for` in `ComponentBuilder` caused by late initialization during the application startup (we wait until all widgets are in place on startup). Registering callbacks later should immediately call the callback because the callbacks stored in the `LATE_INIT` static variable are only processed on startup and then never again.

Additionally, I added support for using method calls as widget initializes in `view!` which was long overdue.

Finally, the example exposes a problem with the shutdown mechanism of components, which will be fixed with #331 for factories and needs another PR for components (ideally after this PR and #331 have landed). In short, the dialogs opened in the example will never be destroyed because the runtime runs as long as the root exists but the runtime also has a reference to the root (basically cyclic references).

euclio (approved):

PR #319, opened 2022-10-20
Improve deps

AaronErhardt:
#### Summary

- Use [multiple locations](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations) to simplify `cargo publish`
- Exclude libpanel until it's compatible with the other gtk-rs crates
- Disable the nightly workflow that has been causing problems for weeks due to a regression in the Rust compiler

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -35,7 +35,7 @@ jobs:
         run: cargo clippy -- --deny warnings
 
       - name: "clippy all features"
-        run: cargo clippy --all-features -- --deny warnings
+        run: cargo clippy --features "all" -- --deny warnings
 
       - name: "test"
         run: xvfb-run --auto-servernum cargo test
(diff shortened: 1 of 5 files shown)


euclio:
In what ways is libpanel incompatible? It will have a 0.1 release soon: https://gitlab.gnome.org/World/Rust/libpanel-rs/-/merge_requests/26

AaronErhardt:
Yes, but recent changes with gtk-rs-core require an update to be able to compile. Until that happens for libpanel, it doesn't make sense to keep it because it will break our CI, break our docs build and won't work anyway when enabled.
As soon as the new release happens, I'd love to enable it again quickly.

euclio (approved):

PR #323, opened 2022-10-21
Have dependabot check the Github Actions

pentamassiv:
#### Summary
Dependabot can check the versions of the Github Actions. This will help keep them updated.
I also could not find any opened PRs from dependabot, even though I created a pull request to update criterion a few days ago. This leads me to believe dependabot is not working as expected. I pointed dependabot to the folders containing the Cargo.toml files, which should fix it.

#### Checklist

- [ ] cargo fmt
- [ ] cargo clippy
- [ ] cargo test
- [ ] updated CHANGES.md

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


AaronErhardt:
AFAIK there's no need to specify the individual crates of the workspace. Instead, dependabot is smart enough to recognise cargo workspaces.

pentamassiv:
But somehow it appears to be broken. I can't find any PRs opened by dependabot and for example it did not open a PR for Criterion 0.4. A [PR](https://github.com/Relm4/Relm4/commit/a435fab3404237042df21e6466f38109a2a6665f) was merged 19 days ago to bump it to the most recent version but the new version was published 44 ago. That means dependabot did not raise a PR for multiple weeks even though it runs daily.

AaronErhardt:
You're right but the issue is not the workspace but apparently the "examples" crate name. I've quickly fixed in this commit: https://github.com/Relm4/Relm4/commit/f9b86e6d71415ae4ce13abc4f5ad5d0f49afdd91

pentamassiv:
Okay, nevermind then. I reverted the commit regarding checking cargo. Would you like the Github Actions to be automatically checked?

PR #330, opened 2022-10-27
core: Add synchronous API for commands

AaronErhardt:
#### Summary

I think it's better to support both sync and async for commands. Especially for commands that are just CPU-bound tasks, this will be quite useful.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


relm4/src/sender.rs
```diff
@@ -124,7 +149,22 @@ impl<C: FactoryComponent> FactoryComponentSender<C> {
...
 
-    /// Spawns a future that will be dropped as soon as the component is shut down.
+    /// Spawns a synchronous command.
+    ///
+    /// This is particularly useful for CPU-intensive background jobs that
+    /// need to run on a thread-pool in the background.
+    ///
+    /// The command will always complete, so it's better to
+    /// call [`try_send`](Sender::try_send) if you expect the component
+    /// to be dropped while the command is running.
```
euclio, line 159: This seems like a potential footgun. Won't this always be a problem because the user could close the application at any time?

AaronErhardt: You're right, this isn't a great solutions.

A solution to this problem could be the following: Offer another sender type (maybe called `TrySender`) that is used only for `CommandOutput` whereas `Input` and `Output` still use the current sender implementation. The idea behind this is that `Input` and `Output` are hardly ever used after a component was destroyed (even if it is possible in theory). Sending a `CommandOutput` on the other side should probably return a `Result` by default so it more obvious that sending is fallible.

euclio: That sounds better to me, and is more in-line with what the stdlib receiver API does.

I'm also not sure that `sync_command` is the best name, since to me it sounds like an API that will wait for the command to complete on the same thread. `blocking_command` isn't a great name either for the same reason.

Maybe we should call this `spawn_command` or something like that?

AaronErhardt: Senders now return a `Result` just like every other channel implementation. Where possible, methods like `input()` or `output()` unwrap the result automatically to greatly reduce the need for unwraps by the user. 

Also, `SharedState` and `Reducer` are now aware that sending may fail if a subscribed component is shut down.

examples/components.rs
```diff
@@ -21,7 +21,7 @@ impl SimpleComponent for Header {
                     set_active: true,
                     connect_toggled[sender] => move |btn| {
                         if btn.is_active() {
-                            sender.output(AppMsg::SetMode(AppMode::View));
+                            sender.output(AppMsg::SetMode(AppMode::View)).unwrap();
```
euclio, line 24: I thought the synchronous APIs won't require an unwrap?

AaronErhardt:
Maybe some additional context for the recent changes: I've realized that only the input sender of the component can always be unwrapped (and therefore doesn't need to return `Result`). The input receiver lives in the runtime of the component so sending should never fail. However, output receiver can simply be dropped by detach and commands are not bound to the lifetime of the component except for their oneshot versions. Factories always process the output to be able to forward to the parent component which make sending outputs from factories safe.

So I've decided to make senders return `Result` by default. Sending inputs (or outputs from factories) is save and is unwrapped by Relm4. Everything else must handle the `Result` because there's no guarantee that the receiver is alive.

euclio:
Makes sense. Though, unwrapping still seems like the wrong behavior to encourage, since if it fails it will crash the program. We might want to provide two variants: an `output` variant that just drops (or logs) the error if the send fails, and a `try_output` that returns a `Result` that lets the caller handle it specially if needed. 

I'm guessing that most callers don't care if the sending fails?

AaronErhardt:
It's a tricky question. I've realized that most components have the only purpose of sending output messages. In other words, it is surely a programming error if you detach them. In this case I think panicking (or some other kind of error) is the desired behavior.

Then there are component like `SimpleComboBox` that have both options: Either you listen to messages or you access its state directly. In this case you want to ignore sending errors.

The same could be said about commands. Sometimes you want to panic because a send error indicates that something weird is happening (for example when a component should have a static lifetime).

I've also thought about adding a method that drops the error (and emits a log message). But then I though, that the returned `Result` already covers all possible use cases. You can have a hard error to prevent accidental misuse of components, you can ignore the error or even print a log message.

I don't think however, that silently dropping the error by default is a good option. But maybe we can add convenience methods for this...

PR #337, opened 2022-11-01
Remove actions rs

pentamassiv:
#### Summary

All Actions from actions-rs seem to no longer be maintained. This also applies to actions-rs/cargo. It can just be replaced by the cargo command. I also fixed a small clippy lint to fix the workflows

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


AaronErhardt (approved):
Thanks!

PR #338, opened 2022-11-01
Move the pre-proccessed doc headers into source

pentamassiv:
#### Summary

As mentioned in https://github.com/Relm4/Relm4/issues/292, in order to have the same content in the docs regardless of where they are built, the preprocessed doc headers need to be moved to the source. I moved the collection of links to the source. There are two downsides

- I am not sure it is possible to add the timestamp of the creation to it so I left it out
- The links point to the nightly versions of the other libraries ([gtk4-rs](https://gtk-rs.org/gtk4-rs/git/docs) | [gtk-rs-core](https://gtk-rs.org/gtk-rs-core/git/docs) | [libadwaita-rs](https://world.pages.gitlab.gnome.org/Rust/libadwaita-rs/git/docs/libadwaita) | [libpanel-rs](https://world.pages.gitlab.gnome.org/Rust/libpanel-rs/git/docs/libpanel))
- The link to the book goes to the stable version https://relm4.org/book/stable/

The reason I pointed the links to the nightly versions of the crates is that I don't know how the comments can be conditionally compiled so that they point to the nightly or the stable version. However there is a popup that warns the users about visiting the nightly version.

Are those the only pre-proccessed doc headers?

### DIFF
--- a/relm4-components/src/lib.rs
+++ b/relm4-components/src/lib.rs
@@ -1,7 +1,7 @@
 //! Collection of reusable and easily configurable components for Relm4.
 //!
 //! The docs are available in two versions.
-//! Use the [stable docs](https://docs.rs/relm4_components) if you want get information about a version that was already published.
+//! Use the [stable docs](https://relm4.org/docs/stable/relm4_components/) if you want get information about a version that was already published.
 //! Visit the [nightly docs](https://relm4.org/docs/next/relm4_components/) if are trying out the newest but possibly unstable version of the crate.
 //!
 //! Docs of related crates:
--- a/relm4-components/src/lib.rs
+++ b/relm4-components/src/lib.rs
@@ -1,4 +1,22 @@
 //! Collection of reusable and easily configurable components for Relm4.
+//!
+//! The docs are available in two versions.
+//! Use the [stable docs](https://docs.rs/relm4_components) if you want get information about a version that was already published.
+//! Visit the [nightly docs](https://relm4.org/docs/next/relm4_components/) if are trying out the newest but possibly unstable version of the crate.
+//!
+//! Docs of related crates:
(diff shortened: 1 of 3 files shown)


AaronErhardt:
You can find the preprocessor code here: https://github.com/Relm4/docs/blob/main/append_info.sh

Timestamps are not important on docs.rs because only released versions are available. I think the link to nightly versions of gtk-rs and other crates is fine (because of the warning popover) and the same goes for the book kindof because it also has a link to the nightly version when you open stable.

At some point, we should improve the versioning of the book so individual version are available...

pentamassiv:
Okay, I think I covered all those.

You said you wanted to keep the nightly docs, right? Do you have a suggestion how we should proceed with https://github.com/Relm4/docs/blob/main/append_info.sh? Should it be removed? Then the above mentioned problems would apply to the nightly docs as well.

AaronErhardt:
Yes, the preprocessing of the docs is going to be discontinued as well as the nightly docs in general, probably. My plan would be to stop building docs with CI and instead use docs.rs only starting from 0.5.
If we need docs of pre-releases, publishing them on crates.io should be the best solution.

This means, we probably don't need to link from docs.rs to somewhere else. Only in the other direction to make users aware that the docs have moved to docs.rs (but that could be done with some preprocessing again).

pentamassiv:
Okay, then I changed the links back to the docs from the website. Once a new version is released and it builds on docs.rs, we can add a deprecation warning for the docs on the website. Then we can change the links to the docs.rs version.

AaronErhardt (approved):
I think we should merge both PRs just before the next beta version is released. Yet, with so many pending PRs that still might take some time.

AaronErhardt:
Thanks!

Issue #314, opened 2022-10-11
Do not report missing `Root` or `init_root` in component macro


euclio:
In the `component_empty` example, we report that the `Root` type and the `init_root` function are missing:

https://github.com/Relm4/Relm4/blob/15d1205255e5ae7f27a09a746fdabe7f9926b3de/relm4-macros/tests/ui/compile-fail/component-empty.stderr#L18-L20

However, these are provided by the macro if they're not present, so we should not ask the user to provide them. Instead, the expansion should always provide values for them, even if there isn't enough information provided to actually determine them. `init_root` can just expand to `fn init_root() -> Self::Root { unimplemented!() }`. The `Root` type is a bit harder: we can't use `!` or `std::convert::Infallible`, because the necessary traits aren't implemented. Instead we could default it to `EmptyRoot` or an arbitrary gtk widget. We must only do this if we are sure that the expansion will fail for another reason, though.

PR #340, opened 2022-11-04
macros: Auto-generate the widgets type name

AaronErhardt:
#### Summary

The widgets type is hardly ever referred to from program code, yet the user has to name it for the code to compile.
Arguably, it's better to use the default name of "{self_ty}Widgets" (so "AppWidgets" for an impl on "App") to remove the need to specify the widgets type.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


relm4-macros/src/util.rs
```diff
@@ -2,7 +2,53 @@ use proc_macro::TokenStream;
 use proc_macro2::Span as Span2;
 use syn::punctuated::Punctuated;
 
-use syn::{Ident, Path, PathArguments, PathSegment};
+use syn::spanned::Spanned;
+use syn::{Ident, ImplItem, ItemImpl, Path, PathArguments, PathSegment, Type, TypePath};
+
+pub(super) fn generate_widgets_type(
+    widgets_ty: Option<Type>,
```
euclio, line 9: This can be determined from the item impl instead of passing it in.

relm4-macros/src/util.rs
```diff
@@ -2,7 +2,53 @@ use proc_macro::TokenStream;
...
+    }
+    // Try to generate the type from the self type name.
+    else if let Type::Path(self_ty) = &*component_impl.self_ty {
+        let (path, impl_item) = self_ty_to_widgets_ty(self_ty);
+        component_impl.items.push(impl_item);
+        Some(path)
+    }
+    // Error: No Widget type found or generated.
+    else {
+        let msg = "No `Widgets` found. Please use `type Widgets = WidgetsName;` to name the widgets type.";
```
euclio, line 25: This error message seems inaccurate, wouldn't it only be caused if `self_ty` isn't a normal path?

We should probably validate that earlier if we don't already, and emit an "invalid type for impl" error there. Then we could just unwrap here.

relm4-macros/tests/builder.rs
```diff
@@ -2,10 +2,10 @@ use gtk::prelude::GtkWindowExt;
 use relm4::{gtk, ComponentParts, ComponentSender, SimpleComponent};
 
 #[derive(Default)]
-struct AppModel;
+struct App;
```
euclio, line 5: I wonder if there's a place to put a "Component style guide" that has naming conventions like this in it.

relm4-macros/tests/ui/compile-fail/bad_impl.stderr
```diff
@@ -24,11 +18,12 @@ warning: unused import: `gtk`
   |
   = note: `#[warn(unused_imports)]` on by default
 
-error[E0046]: not all trait items implemented, missing: `Root`, `init_root`
+error[E0046]: not all trait items implemented, missing: `Root`, `Widgets`, `init_root`
```
euclio, line 21: Can we still emit `Widgets` even if the rest of the expansion will fail? See #314.

relm4-macros/tests/component-missing-widgets.rs
```diff
@@ -13,9 +13,9 @@ impl SimpleComponent for TestComponent {
     }
 
     fn init(
-        init: Self::Init,
+        _: Self::Init,
         root: &Self::Root,
-        sender: ComponentSender<Self>,
+        _sender: ComponentSender<Self>,
     ) -> ComponentParts<Self> {
```
euclio, line 19: This could also be a `compile-pass` test, maybe we should start a separate folder for those?

relm4-macros/tests/simple.rs
```diff
@@ -69,10 +68,3 @@ impl SimpleComponent for AppModel {
         }
     }
 }
-
-fn assert_impls_debug<T: std::fmt::Debug>() {}
-
-#[test]
```
euclio, line 75: Why remove this test?

euclio (changes requested):

relm4-macros/tests/component-missing-widgets.rs
AaronErhardt: I also thought about this, but then I realized that every regular test case is already a compile-pass test technically.

relm4-macros/tests/simple.rs
AaronErhardt: The debug impl assert should only be part of one test, I think. It doesn't really make sense to check this several times in different test cases. So I've added this to the simple test again and removed it from the other two test cases which used it (due to copy-pasting the simple example).

relm4-macros/src/util.rs
AaronErhardt: This is true, but then the `widgets_ty` field of the visitors is not needed anymore. What do you think? Should we just remove this from the visitor and handle it in this function instead?

relm4-macros/src/util.rs
AaronErhardt: Actually, the type of `Self` doesn't need to be a path: https://doc.rust-lang.org/reference/items/implementations.html#implementations

I've improved the error message to reflect the actual causes for the error. But apart from that, this is kind of a best effort strategy. As long as a path is used for `Self`, we can generate a `Widgets` type but for any other type of type (which is going to be extremely rare), we don't know what to do and leave the naming up to the user.

relm4-macros/tests/builder.rs
AaronErhardt: That's certainly possible, but requires us to create a custom `ItemImpl` because syn doesn't even parse this: https://github.com/Relm4/Relm4/blob/main/relm4-macros/src/lib.rs#L173

We cannot generate anything if we don't have an `ItemImpl` to place it inside.

AaronErhardt: We could add something to the book. But I don't know whether that is actually necessary. It's only that the examples and tests should be consistent.

relm4-macros/src/util.rs
euclio: Good point. I don't feel strongly either way. Keeping it as-is is fine by me.

relm4-macros/tests/builder.rs
euclio: Agreed. I'm just thinking back to when I started with Relm4 and didn't know what to call anything :)

relm4-macros/tests/component-missing-widgets.rs
euclio: Personally I would move the regular tests to compile-pass tests then. There we get the additional checking that no warnings are emitted.

relm4-macros/tests/simple.rs
euclio: Sounds good.

relm4-macros/tests/ui/compile-fail/bad_impl.stderr
euclio: I see. This specific case doesn't seem worth improving.

euclio (approved):

relm4-macros/tests/component-missing-widgets.rs
AaronErhardt: I'll merge this PR for now, but we can address this in another PR if you want.

PR #344, opened 2022-11-05
Fix public template children

AaronErhardt:
#### Summary

Public templates now have public template children and don't generate unnecessary local variables in the `update_view()` function.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


relm4-macros/src/widgets/gen/util/widget.rs
```diff
@@ -18,11 +18,13 @@ impl ViewWidgets {
 
 impl Widget {
     // Find all template children and get their variables in scope.
-    pub(crate) fn get_template_child_in_scope(&self, stream: &mut TokenStream2) {
+    pub(crate) fn get_template_child_in_scope(&self, stream: &mut TokenStream2, init: bool) {
```
euclio, line 21: I'd probably prefer an enum over a boolean for this, it's unclear what's going on from the call-site.

AaronErhardt: I've added an enum now :)

euclio (approved):

Issue #342, opened 2022-11-05
Asynchronous Components


edfloreshz:
I'm working on an app that relies heavily on asynchronous operations, can that be done in Relm4?

Perhaps a new trait called `SimpleComponentAsync` could be implemented with async functions?

I'm not really sure if this is possible, therefore the issue.

AaronErhardt:
Actually, that might be possible, I will look into that. But on the other side, components already have commands which give you pretty solid async support as well. 

Btw. there's an open PR that allows you to transform a component into a message stream with `StreamExt` etc.: https://github.com/Relm4/Relm4/pull/311

edfloreshz:
I'll try to see if the PR solves my use case, thanks! 

Still, I think this is worth looking into, would be nice to be able to call async without using commands. 

Let me know if I can be of use.

PR #345, opened 2022-11-05
core: Add asynchronous components

AaronErhardt:
#### Summary

Add asynchronous components that run on the GLib main context.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


edfloreshz:
Can this be applied to Factory Components as well?

edfloreshz:
Closes #342

edfloreshz (approved):
The changes look good to me, I tested them and everything looks fine. 

Great work 🎉

edfloreshz:
I just realized that factories don't render anything inside async components.

AaronErhardt:
> I just realized that factories don't render anything inside async components.

Can you provide an example for that?

edfloreshz:
I was using a factory inside my component, once I turned into an async one the factory was no longer displayed. 

You can try to use the same factory in a sync and async component.

AaronErhardt:
Ok, I'll try to modify the factory example to reproduce this.

edfloreshz:
Okay, let me know if you can replicate this.

AaronErhardt:
No, can't replicate. Switching to `SimpleAsyncComponent` in the factory example works just like the regular version.

edfloreshz:
That's strange, I might be missing something then. In that case, I think it's safe to merge.

AaronErhardt:
> That's strange, I might be missing something then. In that case, I think it's safe to merge.

I agree, let's merge this. We already have enough open PRs :sweat_smile: 
Also, factories and components might need some clean-ups anyway after the dust has settled on all the things happening currently.

PR #348, opened 2022-11-12
core: Rework component shutdown

AaronErhardt:
#### Summary

Successor of #152. Similar to #331, but for components.

The idea in the initial approach to configurable components stopping was to let the user decide whether they want to bind the runtime of the component to the `Controller` or the root widget. The problem with that is that the root widget was never a real option because it's a smart pointer in most cases. Binding the lifetime of the runtime to the lifetime of the root widget creates a cyclic reference. And this cyclic reference was the default behavior so far (so components always had static lifetime).

This PR addresses the issue by binding the lifetime to the `Controller` instead. If you still want static components even after dropping a `Controller` you can use `detach_runtime()`.

Also, there apparently was a big problem with dropping senders (which typically happens when the `Controller` is dropped): Once the sender was dropped, the runtime would be stuck in an infinite loop because receivers just immediately return `Poll::Ready` when all senders were dropped.

Additionally, there are several clean-ups in this PR.

~~Most problems should be addressed already, but there's still a weird test failure that must be addressed eventually.~~

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


relm4/src/component/shutdown_on_drop.rs
```diff
@@ -3,29 +3,29 @@ use tokio::sync::oneshot;
 /// A type that destroys an [`AsyncComponent`](crate::async_component::AsyncComponent)
 /// as soon as it is dropped.
 #[derive(Debug)]
-pub(crate) struct DestroyOnDrop {
+pub(crate) struct ShutdownOnDrop {
```
euclio, line 6: I'd move this to `component/mod.rs` or at least re-export it there to avoid needing to use this module name elsewhere.

euclio (approved):
One small suggestion, but otherwise looks good!

PR #350, opened 2022-11-13
core: Allow temporary init of async components

AaronErhardt:
#### Summary

Technically, async components can already have temporary widget initialization through `init_root()`. Yet, this method is used by the macro and changing it would add additional complexity to the macro. So instead, I suggest adding a `temporary_init()` method that allows you to add some widgets as placeholder (for example a loading animation) until the async `init()` method resolves.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


relm4/src/component/async/traits.rs
```diff
@@ -38,6 +38,13 @@ pub trait AsyncComponent: Sized + 'static {
     /// Initializes the root widget
     fn init_root() -> Self::Root;
 
+    /// Allows you to initialize the root widget with a temporary value
+    /// as a placeholder until the [`init()`](AsyncComponent::init)
+    /// future completes.
+    ///
+    /// This method does nothing by default.
+    fn temporary_init(_root: &mut Self::Root) {}
```
euclio, line 46: This is a cool idea! I don't really like the name though. "temporary" doesnt' really imply "loading" for me.

What do you think about `init_loading`? Or `init_placeholder`?

AaronErhardt: I like your suggestions. Maybe `init_loading_widgets` would be even better?

euclio: Works for me.

AaronErhardt: Done

euclio (approved):

PR #352, opened 2022-11-13
core: Add async factories

### LINKED ISSUE #346: Asynchronous Factory Components
Since `AsyncComponent`s can have factory components in them, it makes sense to have asynchronous factory components as well.

AaronErhardt:
#### Summary

@edfloreshz it's finally here! This is so incredible cool when you see it in action!

This PR is mostly complete but still needs some clean-ups here and there, but I need some sleep now so I'll do that later.

 Fixes #346

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


edfloreshz:
Everything has been working fine so far. 🎉

edfloreshz (approved):
Code looks good! 🎉 

Could it be possible for Relm4 to automatically remove the loading widgets and render `view!` once `init_model` is done?

AaronErhardt:
> Could it be possible for Relm4 to automatically remove the loading widgets and render view! once init_model is done?

Well, we could use [RelmRemoveExt](https://relm4.org/docs/next/relm4/trait.RelmRemoveExt.html) to remove the widget again. But then the root has to implement this trait.

Also, the current solution isn't ideal for non-container widgets (such as Labels) in general, because you can't replace the root. On the other side, we need a container anyway for making this possible and "forcing" people to use a Box as root even if there's just a Label once things have loaded that'd be fine I guess.

I'll merge this now, but we can discuss this on Matrix or a separate issue.

PR #353, opened 2022-11-17
core: Draft to improve removal of loading widgets

AaronErhardt:
#### Summary

Add a `TempWidgets` type that takes care of removing loading widgets from a container automatically.
Async components and factories can use this type with the `init_loading_widgets()` method to automatize the removal.

The advantages of this approach are:

+ No (visible) generics (by using `dyn` for most things)
+ Not restricted to the root widget
+ No further trait bounds or associated types
+ A good default implementation that does nothing

#### ToDo

- [x] Give `TempWidgets` a better name, for example `LoadingWidgets`
- [x] Add docs
- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


relm4/src/temp_widgets.rs
```diff
@@ -0,0 +1,91 @@
...
+    children: Vec<C::Child>,
+}
+
+impl<C: RelmRemoveExt> RemoveTempChild for TempWidgetsInner<C>
+where
+    C::Child: AsRef<C::Child>,
+{
+    fn remove(&mut self) {
+        for child in &mut self.children {
+            //let adaptor: ChildRefAdaptor<C> = ChildRefAdaptor { child };
```
euclio, line 18: Remove commented-out code

relm4/src/factory/async/builder.rs
```diff
@@ -109,14 +111,15 @@ where
 
             crate::spawn_local(async move {
                 let data = C::init_model(init, &index, component_sender).await;
+                if let Some(loading_widgets) = loading_widgets {
```
euclio, line 114: Have you considered removing on drop instead?

AaronErhardt:
All points should be addressed, I also went through the checklist and added everything else needed to pass the CI pipeline.

euclio (approved):

PR #354, opened 2022-11-18
macros: Insert types and fns into broken item impl

AaronErhardt:
#### Summary

Fixes (or at least improves) the situation regarding #314

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


euclio:
This doesn't look like it fixes the intended issue. I would expect [these lines](https://github.com/Relm4/Relm4/blob/main/relm4-macros/tests/ui/compile-fail/component-empty.stderr#L18-L19) to be removed from the output.

PR #355, opened 2022-11-18
core: Pass root to update and update_cmd

AaronErhardt:
#### Summary

Currently, the root is only passed to the `init()` function but not available later. Therefore, if you want to create a dialog as response to a message, you can't simply call `set_transient_for()`. As a workaround, you can use `update_with_view()` which isn't a great solution.

This PR adds the root as a parameter of `update()` and `update_cmd()` for `Component` and `AsyncComponent`.
All sub-traits like `SimpleComponent` are not affected.
This makes it easier to use dialogs the way they are intended to be used by the GTK developers (a window that waits for a response and is destroyed afterwards).

The changes in the "message_stream" example show how this helps with creating dialogs.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@
 ### Changed
 
 + core: Add `gnome_42` and `gnome_43` feature flags
++ core: Pass `Root` during `update` and `update_cmd` for `Component` and `AsyncComponent`
 + core: Rename `OnDestroy` to `RelmObjectExt`
 + core: Remove `EmptyRoot` in favor of the unit type
 + macros: Allow using methods calls as widget initializers in the `view` macro
--- a/examples/drawing.rs
+++ b/examples/drawing.rs
@@ -66,7 +66,7 @@ impl Component for App {
       }
     }
 
-    fn update(&mut self, msg: Msg, _sender: ComponentSender<Self>) {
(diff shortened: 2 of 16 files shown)


AaronErhardt:
In the end, this might be a question of preference, but I don't like using `update_with_view`. The whole point of using the Elm architecture is to separate UI from application logic and I don't want to mix them again just because I need a workaround for spawning dialogs...

euclio:
Could we add a getter `root()` to the trait instead?

AaronErhardt:
Adding a getter for the root to the trait isn't going to help according to my understanding. You still have to store the root somewhere which would be the model in this case (you could already do that but that's quite ugly IMO).

The only thing I was thinking about lately was adding the root the component sender types. I thought that wouldn't work because the component sender types should be `Send`able and storing the root would make it non-sendable, but we could use `Fragile` to work around that. Sending the component sender to another thread is very rare and trying to access the root from another thread is even more unlikely. And in the worst case, accessing the root from another thread would cause a panic.
So maybe we add a `root()` getter to all the component sender types (sync + async + factory senders)? That would even work for the simplified trait variants like `SimpleComponent`.

euclio:
Ah, my mistake. You're right. I prefer this solution to adding to the sender, I think that would be a confusing API.

euclio (approved):

PR #372, opened 2022-12-09
retry add clone trait to factoryVecDeque

sashinexists:
#### Checklist

- [ ] cargo fmt
- [ ] cargo clippy
- [ ] cargo test
- [ ] updated CHANGES.md

### DIFF
--- a/relm4/src/factory/sync/collections/vec_deque.rs
+++ b/relm4/src/factory/sync/collections/vec_deque.rs
@@ -580,9 +580,13 @@ impl<C: FactoryComponent> FactoryVecDeque<C> {
     }
 }
 
-impl<T: Clone + FactoryComponent> Clone for FactoryVecDeque<T>
+trait CloneableFactoryComponent: FactoryComponent {
+    fn get_init(&self) -> Self::Init;
+}
+
+impl<C: FactoryComponent> Clone for FactoryVecDeque<C>
 where
-    T::Init: From<T>,
+    C: CloneableFactoryComponent,
 {
     fn clone(&self) -> Self {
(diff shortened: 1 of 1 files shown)


relm4/src/factory/sync/collections/vec_deque.rs
```diff
@@ -579,3 +579,21 @@ impl<C: FactoryComponent> FactoryVecDeque<C> {
         output
     }
 }
+
+impl<T: Clone + FactoryComponent> Clone for FactoryVecDeque<T>
+where
+    T::Init: From<T>,
+{
```
AaronErhardt, line 586: Looks good overall, but I wonder whether it'd be better to use `T::Init: From<&T>` and remove the requirement `T: Clone` entirely. You might not want to clone the entire data when you just need `Init` in the end.

relm4/src/factory/sync/collections/vec_deque.rs
```diff
@@ -579,3 +579,27 @@ impl<C: FactoryComponent> FactoryVecDeque<C> {
         output
     }
 }
+
+/// Adds a get_init function to FactoryComponents that are clonable
+trait CloneableFactoryComponent: FactoryComponent {
+    fn get_init(&self) -> Self::Init;
+}
```
AaronErhardt, line 586: Looks good overall, but this trait must be public, otherwise nobody will be able to implement it. I suggest we move it to `relm4/src/factory/sync/traits.rs` where it can be located much easier.

AaronErhardt (changes requested):

PR #379, opened 2022-12-24
Prevent leaking CommandSenderInner struct

Ayush1325:
#### Summary
Earlier, this struct was being leaked due to the implementation of Deref. However, since the struct was never actually exported, documentation was not being generated for it's methods.

This commit just makes the struct private and instead creates wrapper functions over it. This is typically whats done in Rust std.

Signed-off-by: Ayush Singh <ayushsingh1325@gmail.com>

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
++ core: Prevent leaking `CommandSenderInner` struct
+
 ## 0.5.0-rc.1
 
 ### Added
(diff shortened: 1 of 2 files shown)


euclio (approved):

AaronErhardt (changes requested):
Looks good. Could you add an entry for this to the changelog in the category "Fixed"?

Ayush1325:
> Looks good. Could you add an entry for this to the changelog in the category "Fixed"?

Should I add a new Release as well, or just add the changes to `0.5.0-rc.1`?

AaronErhardt:
Just put a `### Fixed` below `## Unreleased `. We collect changes under Unreleased until a new version comes out.

AaronErhardt (approved):
Thanks!

PR #385, opened 2023-01-14
Relm4 format

AaronErhardt:
#### Summary

+ Adds relm4-format as CLI tool for formatting code in the view macro
+ Moves internal macro code into another crate to allow reusing the code of the view macro for the formatter (proc-macros can't be used as regular dependency)

The tool can be installed with `cargo install --path=PATH_TO_RELM4_FORMAT_CRATE`. 
It can be called from the command line with `relm4-format file1 file2` or just `relm4-format` to format all Rust files contained in the working directory. Be cautious, commit all changes before using!

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [ ] updated CHANGES.md


euclio:
Would be nice to handle GTK macros like `clone!` too.

relm4-macros/relm4-macros-internal/src/widgets/parse/view_widgets.rs
```diff
@@ -13,6 +13,12 @@ impl Parse for ViewWidgets {
 
         while input.peek(Token![,]) {
             let _colon: Token![,] = input.parse()?;
+
+            // Allow ending with a comma after the last widget
```
PJungkamp, line 17: This should really be it's own commit and probably a separate PR which could be merged immediately. Trailing comma support on the widgets is one of the rough edges of relm4 that I've encountered too often.

PJungkamp:
I haven't given this a proper review as I'm not familiar with the internals of relm4 yet. But I want to add some comments based on a quick scan of the code.

- Why are the `relm4-macros-internal` and `relm4-format` crates nested in the `relm4-macros` directory? I find this quite unusual for a workspace. I think a having something like `relm4-macros`, `relm4-macros2` and `relm4-format` all top-level next to the other workspace members might make more sense.
- Adding conditional compilation with `#[cfg(feature = format)]` for the `name_assigned_by_user` probably isn't necessary. I think this attribute might also come in handy for fixing my issue https://github.com/Relm4/Relm4/issues/532, where the `view!` macro uses a call site hygiene in places where mixed hygiene would be more appropriate.

AaronErhardt:
Thanks for your review!
This likely isn't going to be in its current form however, because parsing files with syn eliminates comments, which makes sense for macros which should ignore them, but makes proper formatting impossible (at least if you don't want to erase comments). It's more likely that - if I find time for this - I will work on an alternative crate to syn which not only could greatly improve parsing and error handling, but also allow parsing comments for formatters like this.

PR #389, opened 2023-01-19
Add method `broadcast` to `FactoryVecDeque`

jirijakes:
#### Summary

In order to send a message to all elements in `FactoryVecDeque`, it was previously necessary to manually cycle through all the indices (from 0 to `len()`) and send message to each element, one by one, using an existing `send` method. This new method makes this process cleaner and safer by avoiding the need to use panic-prone indices.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
++ core: Add `broadcast` to `FactoryVecDeque`
+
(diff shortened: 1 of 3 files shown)


AaronErhardt (approved):
Thanks, looks good!

PR #390, opened 2023-01-20
Fix a clippy warning caused by the macro's codegen

AaronErhardt:
#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


relm4-macros/src/token_streams.rs
```diff
@@ -144,7 +144,9 @@ impl Widget {
         if generate_init_root_stream {
             if let Some(root_name) = root_name {
                 streams.rename_root.extend(quote_spanned! {
-                    name_span => let #name = #root_name.clone();
```
euclio, line 147: Could the issue be caused by the fact that the expanded tokens use the span of the name here? Clippy shouldn't warn if the span of the warning is part of the macro.

AaronErhardt: True, good catch. It can be tempting to use use attributes, but sometimes the fix is even easier. Most importantly, we now have a test case though that ensures we keep this edge case covered.

euclio (approved):

PR #393, opened 2023-02-03
core: Fix crash caused by UID overflow with very large or frequently changing factories

Evilhotdog:
#### Summary

Previously, the UID of a factory collection could overflow when there had been enough changes to a `FactoryVecDeque` to go over the 16-bit integer limit. The type of the UID has been bumped from a `u16` to a `usize`, and every time `clear` is called, the UID is reset to 1. 

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


AaronErhardt (approved):
Awesome, thanks!

PR #403, opened 2023-02-09
macros: Use type fullname

sanpii:
#### Summary

The code generated by macro requires importing `ComponentSender`:

```
error[E0412]: cannot find type `ComponentSender` in this scope
  --> src/application/widget.rs:14:1
   |
14 | #[relm4::component(pub)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
   |
   = note: this error originates in the attribute macro `relm4::component` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this struct
   |
1  | use relm4::ComponentSender;
   |

For more information about this error, try `rustc --explain E0412`.
```

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [ ] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@
 + core: Fix scaping of examples on docs.rs
 + core: Fix crash caused by UID overflow with very large or frequently changing factories
 + macros: Fix clippy warning triggered by the view macro in some edge cases
++ macros: Import `relm4::ComponentSender` isn’t longer required
 
 ## 0.5.0-rc.1 - 2022-12-21
 
--- a/relm4-macros/src/component/mod.rs
+++ b/relm4-macros/src/component/mod.rs
@@ -123,10 +123,10 @@ pub(crate) fn generate_tokens(
             ..
         } = PreAndPostView::extract(&mut component_impl, errors);
 
-        let sender_ty: syn::Ident = if asyncness.is_some() {
-            parse_quote! { AsyncComponentSender }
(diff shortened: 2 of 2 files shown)


AaronErhardt (changes requested):
Looks good, thanks!

Apparently, the factory macro suffers from the same problem: https://github.com/Relm4/Relm4/blob/main/relm4-macros/src/factory/mod.rs#L105-L109
Could you fix that, too?

Also, relm4-macros has removed root imports a couple of months ago because there was no practical advantage and crates that re-export Relm4 suffered from that. Therefore the convention is to just use `relm4::PathToType` for imports.

sanpii:
> Apparently, the factory macro suffers from the same problem: https://github.com/Relm4/Relm4/blob/main/relm4-macros/src/factory/mod.rs#L105-L109 Could you fix that, too?

Done.

> Also, relm4-macros has removed root imports a couple of months ago because there was no practical advantage and crates that re-export Relm4 suffered from that. Therefore the convention is to just use `relm4::PathToType` for imports.

Ok, It’s changed.

AaronErhardt:
Awesome! Could you add a short message to the changelog?

sanpii:
I rollbacked the modifications for the factory macro, this is already done:
https://github.com/Relm4/Relm4/blob/main/relm4-macros/src/factory/mod.rs#L120-L120

I updated the changelog.

AaronErhardt (approved):
Thanks!

Issue #411, opened 2023-02-18
RelmApp API for using MessageBroker with the top-level Component


azymohliad:
Hey! So [as we discussed in Matrix room](https://matrix.to/#/!DlltQISlgRGacjtFQD:matrix.org/$pJkYD1Sle6P9PZHF9dT_E6DkErlTQUiImOvb8ZVe2rs?via=matrix.org&via=gnome.org&via=the-apothecary.club), I'm trying to add builder pattern to `RelmApp` to allow message broker to be attached to the top-level component.

I guess there's no much point to use proper builder pattern with `RelmAppBuilder` here versus adding modifier methods to `RelmApp` because it behaves like a builder itself: simply stores data until consumed when run. Ideally, I hoped it would allow it to use it like this:

```rust
RelmApp::new<Model>(app_id)
    .with_args(&some_args)
    .with_broker(&some_broker)
    .as_async(true)
    .run()
```

But there are a few issues here.

First, to store those optional fields `RelmApp` would need to be generic over component but also over args type, which for sync component only could look like this:

```rust
pub struct RelmApp<'broker, 'args, C: Component, S: AsRef<str>> {
    app: gtk::Application,
    broker: Option<&'broker MessageBroker<C>>,
    args: Option<&'args [S]>,
}
```

And when `with_broker` and `with_args` aren't used (and I assume this is most of the time), Rust would require to specify those generic types explicitly, which doesn't make much sense when those aren't used.

Second, it seems to be quite hard and probably unreasonable to make it properly generic over both sync and async components, and it looks easier to stick to separate `run<C: Component>()` and `run_async<C: AsyncComponent>()` methods. MessageBroker is implemented for sync only, but I assume it should be easy to add it to async too. It looks like `MessageBroker` doesn't use anything besides `Input` type from the `Component` it is generalized over, so maybe it can rather be generalized over message type? And then if the same message broker struct could be reused for async, then I guess `RelmApp` could be generic over message type and store the broker.

But overall, I'm really not sure how to proceed. Just adding `run_with_broker` and `run_with_args_and_broker` methods doesn't seem like such a bad idea anymore, but only until message broker is implemented for async.

What are your thoughts?

AaronErhardt:
I'll have a closer look at this tomorrow. Maybe using `dyn` or even converting it to `&str` directly would at least help with getting rid of generics for the args and the additional lifetime.

azymohliad:
[...]

PR #412, opened 2023-02-18
Add builder methods to RelmApp to allow registering top-level MessageBroker

azymohliad:
#### Summary

Adds builder methods to `RelmApp` which makes it possible to use `MessageBroker` with top-level `Component`.
Addresses #411

Worth noting:
- To make it possible I changed `MessageBroker` generic from `Component` to message type itself.
- I also renamed `RelmApp::with_app` to `RelmApp::from_app` to avoid confusion for builder modifiers, such as `with_args` and `with_broker`.
- I marked `run_with_args` and `run_async_with_args` as deprecated, as they seem unneeded now.
- The args are stored in `RelmApp` as `Vec<String>` rather than generic, otherwise the user would have to specify the type explicitly when not using `with_args`, which would be very inconvenient. But of course this brings some overhead at runtime, so I'm not sure if it's the best way.
- I made `run_async` panic if the broker is set, as launching with message broker isn't implemented for async component (yet?).

I will update changelog after the discussion, because this is just a draft for now. 

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


AaronErhardt:
Looks very good in general. The problem is however, that this is a breaking change according to semver, so we can't release this in the 0.5 series. So either we adjust the API to avoid breaking changes or we target 0.6, which for now lives on the "next" branch and should be deemed stable in roughly two or three months.

azymohliad:
Thanks for a quick review! Yeah, I assumed it would target 0.6.

I don't really see any good way to keep it concise and fully backward compatible. Renaming `with_app` is avoidable. But the main problem is `MessageBroker` generic type parameter. If we keep it as  `MessageBroker<C: Component>` then `RelmApp` struct would need to be bound to `C: Component` too, so it won't be usable with `AsyncComponent`.

I guess the decision also depends on if it's generally the right direction to change generic type parameter in `MessageBroker`. Are there any other concerns about it? Was it supposed to use anything else from `Component` besides the `Input` associated type? Also, what was the plan for message broker with async components? Can the same `MessageBroker` struct be reused for both? Or was it put under `component::sync` module because there's something fundamentally incompatible with async components, and should there be a separate message broker implementation in `component::async`? (in which case the current state of this PR won't work for async components anyway)

AaronErhardt:
I think the only option for backwards compatibility would be to introduce a new `RelmAppBuilder` type and even that would have limitations. 

`MessageBroker` should only need the input message actually. This PR could indirectly clear the way for an async variant of `MessageBroker` actually. Or maybe we add that first to avoid running into problems later?

As for compatibility between async and sync variants, the reason why there are so many different implementations and code duplication is Rust's type system. I would love to have a shared sub-trait between `Component` and `AsyncComponent` but that won't work until specialization lands in Rust. Currently, the compiler still complains that you technically could implement both traits for the same type which would give you two implementations of the sub-trait.
TLDR; something like `AsyncMessageBroker` should be easy to implement but sadly might need some code duplication with only subtle modifications

azymohliad:
> I think the only option for backwards compatibility would be to introduce a new `RelmAppBuilder` type and even that would have limitations.

Hm.. I might be missing something (quite likely, I'm a bit sleepy and probably should reply tomorrow), but I don't see how `RelmAppBuilder` would help to fully preserve the backward compatibility :thinking: Even if `RelmApp` is constructed by `RelmAppBuilder`, it would still need to store `MessageBroker` as its field so that it could use it in `run` method. So if `MessageBroker` type parameters remain unchanged for backward compatibility, `RelmApp` would have to be bound to sync component as its generic type parameter:

```rust
pub struct RelmApp<C: Component> {
    // ...
    broker: Option<&'static MessageBroker<C>>,
}
```

> `MessageBroker` should only need the input message actually. This PR could indirectly clear the way for an async variant of `MessageBroker` actually. Or maybe we add that first to avoid running into problems later?

Sure, I don't mind to wait with this PR.

azymohliad:
> I think the only option for backwards compatibility would be to introduce a new `RelmAppBuilder` type and even that would have limitations.

Hm.. I very likely may be missing something, but I don't yet understand how `RelmAppBuilder` could help with backward compatibility :thinking: Even if `RelmApp` is constructed by `RelmAppBuilder`, it would still need to store `MessageBroker` as its field so that it could use it from the `run` method. So if `MessageBroker` generic type parameter remains unchanged for backward compatibility, `RelmApp` would have to be bound to sync component:

```rust
pub struct RelmApp<C: Component> {
    // ...
    broker: Option<&'static MessageBroker<C>>,
}
``` 

Or are you more concerned about `with_app` -> `from_app` rename here?


> `MessageBroker` should only need the input message actually. This PR could indirectly clear the way for an async variant of `MessageBroker` actually. Or maybe we add that first to avoid running into problems later?

Sure, I don't mind to wait with this PR.

Thanks for detailed explanation. But so if I understand correctly, with the changes here there's probably no need for duplicated `AsyncMessageBroker` implementation, and maybe the current one can be used from both (and maybe moved to something like `component/common/message_broker.rs`), right?

AaronErhardt:
The idea with backwards compatibility only works if the builder doesn't interact with `RelmApp` at all and we also need a sync and an async variant of the builder. Instead of returning a `RelmApp` you'd just finish the builder with `.run()` or something similar. I might have missed something myself though...

> Thanks for detailed explanation. But so if I understand correctly, with the changes here there's probably no need for duplicated AsyncMessageBroker implementation, and maybe the current one can be used from both (and maybe moved to something like component/common/message_broker.rs), right? 

That could be very much possible. But then things could also get complicated internally. I'll look at this tomorrow.

AaronErhardt:
After looking at the code, I think it should be easy to add a `launch_with_broker()` method to `AsyncComponentBuilder`.

You should be able to just copy this to the `relm4/src/component/async/builder.rs:144`:

```rust
    /// Similar to [`launch()`](AsyncComponentBuilder::launch) but also initializes a [`MessageBroker`].
    ///
    /// # Panics
    ///
    /// This method panics if the message broker was already initialized in another launch.
    pub fn launch_with_broker(self, payload: C::Init, broker: &MessageBroker<C>) -> AsyncConnector<C> {
    pub fn launch_with_broker(
        self,
        payload: C::Init,
        broker: &MessageBroker<C::Input>,
    ) -> Connector<C> {
        let (input_sender, input_receiver) = broker.get_channel();
        self.launch_with_input_channel(
            payload,
            input_sender,
            input_receiver.expect("Message broker launched multiple times"),
        )
    }
```

Could you try whether this works? Also, you might want to change the PR target branch to next.

azymohliad:
Sure, I'll try it later today. Would you prefer me to add it to this PR as well? Should I finalize it if everything works (update changelog, mark as ready)?

AaronErhardt:
Yes, it would make sense to add this to this PR and then finalize it.

azymohliad:
So it's done and everything seems to work well. I moved `message_broker` from `component::sync` to `component` so that `component::async` would have access to `MessageBroker::get_channel` private method used in `launch_with_broker`. Plus such structure seems more logical anyway, now that `MessageBroker` is used for both.

AaronErhardt (approved):
Looks good to me, very nice job! Thanks!

PR #419, opened 2023-02-25
macros: Fix ordering in codegen

### LINKED ISSUE #418: 5.0 macro property reordering
Take this example:

```toml
# Cargo.toml
relm4 = { version = "0.5.0", features = ["macros"] }
```

```rust
use gtk::prelude::{BoxExt, ButtonExt, GtkWindowExt, OrientableExt, PopoverExt};
use relm4::{gtk, ComponentParts, ComponentSender, RelmApp, RelmWidgetExt, SimpleComponent};

struct AppModel {
    counter: u8,
    max: u8,
}

#[derive(Debug)]
enum AppMsg {
    Increment,
    Decrement,
    ChangeMax(u8),
}

#[relm4::component]
impl SimpleComponent for AppModel {
    type Init = AppModel;
    type Input = AppMsg;
    type Output = ();

    view! {
        gtk::ApplicationWindow {
            set_title: Some("Simple app"),
            set_default_width: 300,
            set_default_height: 100,


            #[wrap(Some)]
            set_titlebar = &gtk::HeaderBar {
                pack_start = &gtk::MenuButton {

                    #[wrap(Some)]
                    set_popover = &gtk::Popover {
                        gtk::Box {

                            gtk::SpinButton {
                                set_increments: (1.0, 5.0),
                                set_range: (4.0, 200.0),
                                set_value: model.max as f64,

                                connect_value_changed[sender] => move |btn|
                                    sender.input(AppMsg::ChangeMax(btn.value() as u8)),
                            },
                        }
                    }
                },
            },


            gtk::Box {
                set_spacing: 5,
                set_margin_all: 5,

                gtk::Button {
                    set_label: "Increment",
                    connect_clicked[sender] => move |_| {
                        sender.input(AppMsg::Increment);
                    }
                },

                gtk::Button::with_label("Decrement") {
                    connect_clicked[sender] => move |_| {
                        sender.input(AppMsg::Decrement);
                    }
                },

                gtk::Label {
                    #[watch]
                    set_label: &format!("Counter: {}", model.counter),
                    set_margin_all: 5,
                }
            }
        }
    }

    fn init(
        model: Self::Init,
        root: &Self::Root,
        sender: ComponentSender<Self>,
    ) -> ComponentParts<Self> {
        let widgets = view_output!();
        ComponentParts { model, widgets }
    }

    fn update(&mut self, msg: Self::Input, _sender: ComponentSender<Self>) {
        dbg!(&msg);
        match msg {
            AppMsg::Increment => {
                self.counter = self.counter.wrapping_add(1) % self.max;
            }
            AppMsg::Decrement => {
                self.counter = self.counter.checked_sub(1).unwrap_or(self.max - 1);
            }
            AppMsg::ChangeMax(max) => {
                self.max = max;
                self.counter %= max;
            }
        }
    }
}

fn main() {
    let app = RelmApp::new("relm4.test.simple");
    app.run::<AppModel>(AppModel {
        counter: 0,
        max: 30,
    });
}

```

Testing it (without any interaction) gives the following: 

```
[src/main.rs:89] &msg = ChangeMax(
    4,
)
[src/main.rs:89] &msg = ChangeMax(
    30,
)
```

In the equivalent code from 4.0, there's no output; no message is fired without interaction. If i set up `connect_value_changed` after `view_output!()`, there's also no output.

If i place `set_value` before `set_range`, only the `ChangeMax(4)` message fires. So for some reason, the `connect_value_changed` is always setup before `set_range`.

I don't think this is the expected behavior, if it is, the [migration guide](https://relm4.org/book/stable/0_4_to_0_5.html) should probably have a warning.

AaronErhardt:
#### Summary

Fixes #418

0.5 brings a new feature with the macro to block signal handlers when assigning properties. This was made possible by generating the signal handlers first and then at each property update with a `#[block_handler(handler_name)]`, the signal handler would be temporarily blocked. 

However, this is not great for initializing properties because even on initialization a signal it fired sometimes unless you use the `block_handler` attribute. With this PR, the old order of code generation is restored: First assign properties, then generate signal handlers. The `block_handler` attribute now only affects property updates done with either `#[watch]` or `#[track]` because initialization won't cause any events anyway.

@imsamuka can you test your code against this branch? I think this should fix it.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
++ macros: Fix order of assignment and connect statements
+
 ## 0.5.0 - 2023-2-17
 
 ### Fixed
--- a/relm4-macros/src/component/mod.rs
+++ b/relm4-macros/src/component/mod.rs
@@ -91,11 +91,11 @@ pub(crate) fn generate_tokens(
             #rename_root
             #menus_stream
             #init_widgets
+            #assign
             #connect
             {
                 #error
             }
-            #assign
         };
 
         let widgets_return_code = parse_quote! {
--- a/relm4-macros/src/factory/mod.rs
+++ b/relm4-macros/src/factory/mod.rs
@@ -88,11 +88,11 @@ pub(crate) fn generate_tokens(
             #rename_root
             #menus_stream
             #init
+            #assign
             #connect
             {
                 #error
             }
-            #assign
         };
 
         let widgets_return_code = parse_quote! {
--- a/relm4-macros/src/view.rs
+++ b/relm4-macros/src/view.rs
@@ -27,8 +27,8 @@ pub(super) fn generate_tokens(input: TokenStream) -> TokenStream {
 
     let output = quote! {
         #init
-        #connect
         #assign
+        #connect
         {
             #error
         }
--- a/relm4-macros/src/widget_template.rs
+++ b/relm4-macros/src/widget_template.rs
@@ -44,8 +44,8 @@ pub(crate) fn generate_tokens(vis: Option<Visibility>, mut item_impl: ItemImpl) 
 
                     let view_output = quote! {
                         #init
-                        #connect
                         #assign
+                        #connect
                         {
                             #error
                         }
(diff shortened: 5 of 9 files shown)


imsamuka:
Seems like it! The example I gave works, and the project I migrated works too.

PR #421, opened 2023-02-27
Fix Stack so that it appears correctly in StackSwitcher, etc

sumibi-yakitori:
#### Summary

I am currently in the process of migrating from relm 0.4 to 0.5 and `Stack` is not displaying correctly.

This behavior does not seem to be documented in GTK4, but it appears that `StackPage` does not appear in `StackSwitcher` or `StackSidebar` unless the title is specified using the `add_titled` method. Therefore, `StackPageInfo` should be used as in relm 0.4.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

Since I am a non-English speaker, please make appropriate changes to the wording I have added to CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
++ core: Fix to use `StackPageInfo` in the implementation of `Stack`'s `FactoryView`
+
 ## 0.5.0 - 2023-2-17
 
 ### Fixed
--- a/relm4/src/factory/positions.rs
+++ b/relm4/src/factory/positions.rs
@@ -22,3 +22,13 @@ pub struct FixedPosition {
     /// Position on the y-axis.
     pub y: f64,
 }
+
+#[derive(Debug, Default)]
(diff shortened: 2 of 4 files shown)


AaronErhardt:
Looks good, but I would really like to avoid such a workaround solution. The returned widget should give you all you need to set the title of the page...
Arguably, the bug is not on Relm4's side. I understand your motivation, but I'm hesitant to merge this, especially since this a breaking change.

sumibi-yakitori:
Hmmm. I certainly feel differently about handling this issue on the part of Relm4.

https://relm4.org/book/stable/component_macro_reference.html?highlight=return#returned-widgets
You are referring to this macro, right?
Does this approach allow for dynamic addition of items using `FactoryVecDeque`?

sumibi-yakitori:
Ah, I finally figured out how to use the `returned_widgets` argument. Thanks!

PR #427, opened 2023-03-15
examples: Add example for relm4-icons

AaronErhardt:
#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


relm4/Cargo.toml
```diff
@@ -56,13 +56,14 @@ relm4-macros = { path = "../relm4-macros" }
...
 # For the examples
-tokio = { version = "1.25", features = [
+relm4-icons = { version = "0.6.0-beta.1", features = ["plus", "minus"]}
+tokio = { version = "1.26", features = [
     "rt",
     "macros",
     "time",
     "rt-multi-thread",
 ] }
-futures = "0.3.26"
```
songww, line 65: The dependencies that have been imported in the 'dependencies' section can be used directly, there is no need to import them again here.

examples/Cargo.toml
```diff
@@ -6,8 +6,9 @@ edition = "2021"
 
 [dev-dependencies]
 relm4 = { path = "../relm4" }
-tokio = { version = "1.25", features = ["rt", "macros", "time", "rt-multi-thread"] }
```
songww, line 9: When running the example with the command `cargo run --example message_stream`, it uses the dependencies defined in relm4/Cargo.toml, but this file is being ignored. Is it necessary to keep this file?

examples/icons.rs
```diff
@@ -0,0 +1,81 @@
...
+                self.counter = self.counter.wrapping_sub(1);
+            }
+        }
+    }
+}
+
+fn main() {
+    let app = RelmApp::new("relm4.example.icons");
+
+    relm4_icons::initialize_icons();
```
songww, line 78: Is it possible to only expose one feature and call this method internally within relm4?

songww (approved):

examples/icons.rs
AaronErhardt: Yes, I thought about this already. The only problem is that you have to include `relm4-icons` as dependency anyway to include some icons via feature flags. So the question is whether having a feature flag or writing one line of code is better...

examples/Cargo.toml
AaronErhardt: I never thought about this but it seems to work. Thanks!

examples/icons.rs
AaronErhardt: I think we should merge this and look for more feedback. After all, we still have a couple of weeks before 0.6 becomes the new stable release.

PR #428, opened 2023-03-16
Export some types as pub for customizing factory types,

songww:
#### Summary
When customizing factory types, these types are very helpful.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [ ] updated CHANGES.md

### DIFF
--- a/relm4/src/factory/async/builder.rs
+++ b/relm4/src/factory/async/builder.rs
@@ -11,7 +11,7 @@ use crate::runtime_util::GuardedReceiver;
 use crate::shutdown::ShutdownSender;
 use crate::{shutdown, Receiver, Sender};
 
-pub(super) struct AsyncFactoryBuilder<C: AsyncFactoryComponent> {
+pub struct AsyncFactoryBuilder<C: AsyncFactoryComponent> {
     init: C::Init,
     pub(super) root_widget: C::Root,
     pub(super) component_sender: AsyncFactorySender<C>,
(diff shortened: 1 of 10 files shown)


AaronErhardt:
> When customizing factory types, these types are very helpful.

You mean something like creating your own data structures? I thought those types were only useful internally, but if you have a use case, could you elaborate shortly why you would need those types?

songww:
If you want to define a factory container type that uses similar usage as FactoryVecDeque, then you need to implement two types similar to ComponentStorage and FactoryBuilder to handle component initialization and access returned_widgets. It would be very convenient if these two types can be reused.

AaronErhardt:
I see, but then again it looks like a shortcoming of Relm4 for me if you have to deal with such low-level stuff. Why exactly doesn't FactoryVecDeque not work for you and what were your problems with customising a factory container? This sounds like something that could be improved in 0.6.

songww:
The windows and opened buffers in neovim all have a numerical identifier, which is used to index their values when modified. I would like to use a type such as VecMap to store these identifiers. Their corresponding widget positions are fixed and laid out by neovim, so I am using gtk::Fixed as their parent container. They do not need to be re-ordered and storing them in random order does not affect me.
When defining this type, I wanted to reuse these auxiliary types, but they are all private. Later on, I encountered DynamicIndex. 
Currently, I am using a VecMap to store the index of the DynamicIndex + FactoryVecDeque for Component storage, but many aspects such as maintaining the DynamicIndex are performance wasteful.

I think other people may also use LinkedList or other types of container to meet their different needs.

songww:
This PR is not very useful, it can be closed.

AaronErhardt:
Sounds like an actual fix for this would be adding FactoryHashMap, right? I was thinking whether the `FactoryComponent` trait is flexible enough for this, maybe we need a new trait for the HashMap approach.

songww:
My other requirement is that neovim sends multiple update events, but these updates can only be applied when the flush event is received.

But now the API applies immediately upon each update.

PR #434, opened 2023-03-22
Improve actions API

### LINKED ISSUE #431: Add a way to install RelmActions at app level
Otherwise, the only way to install actions at app level is the classic GIO API, and this API is not compatible with the `menu!` macro.

AaronErhardt:
#### Summary

Should fix #431

@nahuelwexd what do you think? I think the new `register_for_main_application` method should work for you.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
 
 + core: Documentation and better support for data bindings
 
+### Changed
+
++ core: Improved `RelmActionGroup` API
+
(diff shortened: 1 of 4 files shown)


nahuelwexd:
My mind was thinking of adding an `into_action` to get the `gio::Action`, but this works as well and results in less boilerplate

LGTM 👍

PR #436, opened 2023-03-26
core: Add FactoryHashMap

AaronErhardt:
#### Summary

Supersedes #428 

@songww can you try this patch for your usecase?

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


relm4/src/factory/sync/collections/hashmap.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+}
+
+/// A container similar to [`HashMap`] that can be used to store
+/// values of type [`FactoryComponent`].
+#[derive(Debug)]
+pub struct FactoryHashMap<K, C: FactoryComponent> {
+    widget: C::ParentWidget,
+    parent_sender: Sender<C::ParentInput>,
+    inner: HashMap<K, FactoryHandle<C>>,
```
songww, line 56: I am being a bit demanding, but it would be really great if different hash types could be used.

songww (commented):
Looks good to me.

relm4/src/factory/sync/collections/hashmap.rs
AaronErhardt: Can you elaborate what you mean with different hash types? I'm not sure if I understand what you mean.

examples/entry.rs
```diff
@@ -27,6 +27,7 @@ impl FactoryComponent for Counter {
     type CommandOutput = ();
     type ParentInput = AppMsg;
     type ParentWidget = gtk::Box;
+    type Index = DynamicIndex;
```
euclio, line 30: Should the macro default this to `DynamicIndex`?

euclio (approved):
Looks good. Not thrilled about adding yet another associated type to the trait but I'm not sure if there's a viable alternative.

relm4/src/factory/sync/collections/hashmap.rs
AaronErhardt: I think you meant different hashers, right? Anyway, I've just added support for that.

examples/entry.rs
AaronErhardt: Good suggestion. Is implemented now.

euclio (approved):

PR #449, opened 2023-04-10
Introducing setting and action safeties

### LINKED ISSUE #439: Alternative to menu! and the Actions API.
I feel that the `menu!` macro and the Relm4 actions API are unnecessarily complex and incomplete. It is possible that with a trait like:

~~~ rust
trait SimpleMenu where Self: IsA<gio::Menu> {
	fn  action(&self, action: &str, label: &str);
	fn section(&self, model: &impl IsA<gio::MenuModel>, label: &str);
	fn submenu(&self, model: &impl IsA<gio::MenuModel>, label: &str);
}

impl SimpleMenu for gio::Menu {
	fn action(&self, action: &str, label: &str) {
		self.append(Some(label), Some(action));
	}
	
	fn section(&self, model: &impl IsA<gio::MenuModel>, label: &str) {
		self.append_section((!label.is_empty()).then_some(label), model);
	}
	
	fn submenu(&self, model: &impl IsA<gio::MenuModel>, label: &str) {
		self.append_submenu(Some(label), model);
	}
}
~~~

Get a result like:

~~~ rust
view! {
	gio::Menu {
		action["Some action")]: "win.some-action",
		
		section["Appearance")] = &gio::Menu {
			action["System theme")]: "win.scheme::System",
			action["Light theme" )]: "win.scheme::Light",
			action["Dark theme"  )]: "win.scheme::Dark",
		},
		
		submenu["Hello"] = &gio::Menu {
			action["World!")]: MY_ACTION_CONSTANT,
			action[&gettext("Translatable"))]: OTHER,
		},
	}
}
~~~

If I want "safety" in the action strings, I can use constants.

The first advantage is to be able to bind menus with any implementation of `gio::Action`, and not only the `gio::SimpleAction` through the `RelmAction` wrapper. For example, actions created with `gio::Settings`.

The second is to be able to connect signals of the actions in the view itself, and thus clone widgets or the sender more easily. For example, the "Appearance" menu action corresponds to the following action:

~~~ rust
view! {
	adw::ApplicationWindow {
		add_action = &settings.create_action("scheme") {
			connect_state_notify[sender] => move |this|
				match this.state().unwrap().str().unwrap() {
					"System" => sender.input(Msg::SetSystemScheme),
					"Light"  => sender.input(Msg::SetLightScheme),
					"Dark"   => sender.input(Msg::SetDarkScheme),
					_ => unreachable!()
				},
	}
}
~~~

ejaa3:
#### Summary

Fixes #439

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


AaronErhardt (changes requested):
The PR looks good in general, good job! Yet, a few things can still be improved.

First of all, I'm not sure whether "safeties" is a term that's commonly used in English. It's not easy to come up with an alternative, but maybe we can split the macro into one that implements safe actions and one that implements safe settings. Then we could call one `safe_actions` and the other `safe_settings`. Let me know what you think.

Also, this PR introduces a lot of traits, ideally we could combine some of them. The [`ActionName`](https://docs.rs/relm4/latest/relm4/actions/traits/trait.ActionName.html) trait of the current version for example works for all types of actions, no matter whether they have state or target or both. I'm not sure whether it is still possible to use such a trait, but it would be worth to try it. This would also remove the need to say "Any type that implements it must not implement OtherTrait" because there would only be one trait.

The last thing is that the file size is too large for my taste. For example, there's no benefit in having the macro in the same file as the traits. And the traits themselves can be grouped into traits for safe actions and settings and traits implemented on other types to interact with those actions and settings.

ejaa3:
I'm also not sure “safeties” is the right word. I got carried away by [Rhai](https://rhai.rs/book/safety), so I'm open to any suggestions because I'm a Spanish speaker and I'm okay with the large file size. I'm going to solve it along with other things.

As for splitting the macro, I do not see it feasible. I think it would be spaghetti code, and it is also possible to have an action that is also a setting (examples: [first], [second]). That's why the `ActionSafety` trait "inherits" from `SettingSafety`, since `gio::Settings` creates its actions with the same name as the setting.

[first]: https://github.com/Relm4/Relm4/blob/7872793cbb1f12264949d02a9c08920a6d430376/examples/menu_and_safeties.rs#L128-L131

[second]: https://github.com/Relm4/Relm4/blob/7872793cbb1f12264949d02a9c08920a6d430376/examples/menu_and_safeties.rs#L298-L305

I actually started out naming the macro `safe_actions!` because I only had the actions in mind, not the settings. But even `safe_actions!` would not be creating real `actions` or a wrapper like the old `RelmAction`, but creating a kind of `safety` to use them from a `gio::Action` or `gio::SimpleAction`.

In any case it is possible to invoke the macro several times, one invocation for the actions, one for the settings, and one for the settings actions. Or use comments like this:

~~~ rust
relm4::safeties! {
	// Actions:
	MyAction(group: "win", name: "my-action")
	
	// Settings:
	@state(param: bool)
	MySetting(name: "my-setting")
	
	// Settings actions:
	@state(param: bool)
	MySettingAction(group: "win", name: "my-setting-action")
}
~~~

As for the traits, I see no way to combine them. Or rather, if I combined them it would be less safe than it is. For example, the method `connect_activate_safe_with_state` for `gio::SimpleAction` requires `T` to implement `ActionSafety`, `WithoutValue`, and `Stateful` as the following safety:

~~~ rust
@state(param: bool)
MyAction(group: "win", name: "my-action")
~~~

If I add a `@value` to it, `connect_activate_safe_with_state` would not compile thanks to `WithoutValue`, but `connect_activate_safe_with_target_and_state` would because it requires `WithValue` instead of `WithoutValue`.

The reason the old `ActionName` “works for everything” is because `RelmAction` has a `new` method for each case (with state, with target value, with both or without). But RelmAction does not support settings, does not allow to connect signals more than once and does not let you connect signals in `view!`.

Now, it is not even necessary to document that implementing one trait should avoid implementing another, because the macro does all the work (it would be nice if Rust supported [negative trait bounds](https://github.com/rust-lang/rust/issues/42721)). But I decided to document that so that it is known what the macro does.

Well, I wonder if updating the code will automatically update the PR hehe.

AaronErhardt:
We could re-export the `safe_extensions` in the prelude as well.

I thought maybe `safe_actions!` is enough as name, even if settings are still included. Or maybe`safe_abstractions!`.

examples/README.md
```diff
@@ -7,6 +7,9 @@ A collection of example apps built with Relm4.
 ```bash
 git clone https://github.com/Relm4/Relm4.git
 cd Relm4
+mkdir -p ~/.local/share/glib-2.0/schemas/
+ln -s $(pwd)/examples/relm4.example.safeties.gschema.xml ~/.local/share/glib-2.0/schemas/
+glib-compile-schemas ~/.local/share/glib-2.0/schemas/
```
AaronErhardt, line 12: I would mention that only the new example needs this. The other examples run fine without this, so we should not bother everyone with this.

ejaa3: And what should I do? `gio::Settings` requires a schema, and the easiest way to pass it one is to have a schema in ~/.local/share/glib-2.0/.

Or do you mean that I should put that code somewhere else, and specify that running the `menu_action_and_settings` example requires running that?

AaronErhardt: > Or do you mean that I should put that code somewhere else, and specify that running the menu_action_and_settings example requires running that?

Exactly. You can keep it in the README, but ideally in a separate block with a comment that it is only necessary for the `menu_action_and_settings` example.

ejaa3:
I think `relm4::safe_settings_or_actions!` is the best choice because it is explicit and takes one line of code for it, so there is a lot of space left over. `safe_abstractions!` is not specific and `safe_actions!` forget the settings. And the "or" is not exclusive.

As for the module, I think it's okay to dedicate a line to import:
`use relm4::safe_settings_and_actions::extensions::*`

With this, for me this PR is ready. 🤔

AaronErhardt:
Do you mean `safe_settings_or_actions!` or `safe_settings_and_actions!` as name for the macro? I think `and` is better, right?

ejaa3:
Well, I also think that `and` is fine because it's the module name. I will update.

PR #471, opened 2023-06-19
Add a separator to the leaflet example

sgued:
#### Summary

The libadwaita sidebar/leaflet example did not include a separator, while most sidebar will want one.
There is however a subtlety when it comes to adding a separator, as it needs to be wrapped in a `LeafletPage` with the `navigatable` property set to false. However `LeafletPage` is created automatically by the leaflet and cannot be created manually.
There is therefore a bit of a subtlety is required to get access to the generated `LeafletPage` and set it to not navigatable.

It took me some time and some help from the GTK-RS channel to get how to do it, I think having it in the example will help. 

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/examples/libadwaita/leaflet_sidebar.rs
+++ b/examples/libadwaita/leaflet_sidebar.rs
@@ -58,9 +58,10 @@ impl SimpleComponent for App {
                     }
                 },
 
-                #[name = "separator"]
-                gtk::Separator {
+                append = &gtk::Separator {
                     set_orientation: gtk::Orientation::Vertical,
+                } -> {
+                    set_navigatable: false,
                 },
 
                 gtk::Box {
@@ -123,10 +124,6 @@ impl SimpleComponent for App {
             .bind_property("folded", &widgets.back_button, "visible")
             .flags(glib::BindingFlags::SYNC_CREATE)
             .build();
-        widgets
-            .leaflet
-            .page(&widgets.separator)
-            .set_navigatable(false);
 
         ComponentParts { model, widgets }
     }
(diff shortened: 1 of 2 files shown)


examples/libadwaita/leaflet_sidebar.rs
```diff
@@ -58,6 +58,11 @@ impl SimpleComponent for App {
                     }
                 },
 
+                #[name = "separator"]
+                gtk::Separator {
+                    set_orientation: gtk::Orientation::Vertical,
+                },
```
AaronErhardt, line 64: I think this would be a cleaner solution:

```suggestion
                append = &gtk::Separator {
                    set_orientation: gtk::Orientation::Vertical,
                } -> {
                    set_navigatable: false,
                },
```

AaronErhardt (changes requested):
Thanks, looks much better visually. I wonder if Libadwaita has changes the styling since we added this example...

examples/libadwaita/leaflet_sidebar.rs
sgued: Thanks!

This is indeed better, I wasn't aware of this syntax of the view macro.

AaronErhardt: Yeah, it's rather late in the book, no worries. It's definitely worth to check out the macro reference in the book or the macro-reference example to get an overview over what's available.

Anyway, thanks for the PR!

PR #474, opened 2023-06-21
docs(readme): add `lemoa`

Bnyro:
#### Summary

Lemoa is a desktop client for Lemmy (which is an open source link aggregator comparable to Reddit).

Thank you for making it possible by providing such an awesome crate Aaron, it's so, so pleasant to work with Relm4!


AaronErhardt (approved):
Wow, that's awesome! I've just started to be more active on Lemmy again a couple of days ago and it's so nice to see that a native GTK client is already in development :rocket: :rocket: :rocket:

PR #480, opened 2023-06-27
ci(macros): Add tests for returned widgets without builders

jaysonsantos:
#### Summary
Hi there, while trying to use returned widgets with a widget generated by a container, I could not find any docs and tried fiddling with the tests (as they are a more up-to date than the docs) to make it work. Maybe they can also work as docs?

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test


AaronErhardt (approved):
Looks good, thanks!

I hope I can find some time in the future to improve the documentation of the macro, but contributions would also be very welcome :)

PR #489, opened 2023-07-16
Clippy fixes

vorot93:


relm4/src/channel/component.rs
```diff
@@ -145,6 +145,7 @@ macro_rules! sender_impl {
                 shutdown: ShutdownReceiver,
             ) -> Self {
                 Self {
+                    #[allow(clippy::all)]
                     shared: Arc::new(ComponentSenderInner {
```
AaronErhardt, line 149: Why is this necessary? I don't get a warning here and even if clippy complained, it doesn't make sense to ignore all clippy lints here.

AaronErhardt (changes requested):
Looks good overall, there's just one thing that needs to be changed.

AaronErhardt:
> Beta clippy complains with a new lint that's not yet in stable. See previous commits of this PR.

I only use stable and nightly on my system so I couldn't reproduce it. It turns out this is a false positive that has been fixed in nightly, but still exists in the beta toolchain: https://github.com/rust-lang/rust-clippy/issues/11076

This means that `#[allow(clippy::all)]` is just a temporary workaround until the fix gets backported to the beta release channel, so I don't this it should be included, even if the CI fails for a couple more days.

AaronErhardt:
Thanks!

PR #490, opened 2023-07-18
Add `set_margin_vertical`/`horizontal` to `RelmWidgetExt`

Skaldebane:
#### Summary
This adds two simple convenience methods to the `RelmWidgetExt` trait with their implementation for `gtk::Widget`.

#### Checklist

- [x] cargo fmt
- [ ] cargo clippy
- [ ] cargo test
- [ ] updated CHANGES.md

PR #491, opened 2023-07-18
Remove unnecessary explicit call to `.into_iter()`

Skaldebane:
#### Summary
Make clippy happy. This causes the `build (beta)` check to fail, cancelling the rest of the builds in https://github.com/Relm4/Relm4/pull/490.

#### Checklist

- [ ] cargo fmt
- [ ] cargo clippy
- [ ] cargo test
- [ ] updated CHANGES.md


Skaldebane:
There are many more issues apparently. Rushing with untested fixes wasn't a good idea after all 😅

PR #524, opened 2023-09-08
core: Add `Init` type to `WidgetTemplate`

AaronErhardt:
#### Summary

Allow passing data to `WidgetTemplate::init()` to make widget templates more flexible.
Due to clever use of the macros, existing code should still compile just fine.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 ### Changed
 
++ core: Add `Init` type to the `WidgetTemplate` trait to allow passing data to `init()` 
 + core: Just require `AsRef<gtk::Window>` in `RelmApp::run`
 + core: Add `AsRef<Root>` as requirement to the `WidgetTemplate` trait
 
--- a/relm4-macros/src/widget_template.rs
+++ b/relm4-macros/src/widget_template.rs
@@ -7,95 +7,146 @@ use crate::{
     widgets::ViewWidgets,
 };
 
-pub(crate) fn generate_tokens(vis: Option<Visibility>, mut item_impl: ItemImpl) -> TokenStream2 {
-    if item_impl.items.len() != 1 {
-        return Error::new(
-            item_impl.span(),
-            "Expected only one view macro and nothing else",
-        )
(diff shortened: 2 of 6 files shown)


eminfedar:
Is there any way to make widget templates able to access `model`?

I'm defining every component which need to watch changes on the model in main view as a `#[template_child]`, because widget templates can't access the main model like `model.some_variable` 

For example this is not working:
```rust
#[relm4::widget_template(pub)]
impl WidgetTemplate for MainPageFooter {
    view! {
        gtk::Box {
            set_vexpand: true,
            set_valign: gtk::Align::Center,

            gtk::Label {
                set_label: model.time_str,
            },
        }
    }
}
```
I need to make it like this:
```rust
#[relm4::widget_template(pub)]
impl WidgetTemplate for MainPageFooter {
    view! {
        gtk::Box {
            set_vexpand: true,
            set_valign: gtk::Align::Center,

            #[name = "lbl_time"]
            gtk::Label {
                set_label: model.time_str,
            },
        }
    }
}

// And the main `init` function:

#[relm4::component]
impl SimpleComponent for App {
// ...
    view! {
        adw::Window {
/// ...
            #[template]
            MainWindow {

                #[template_child]
                lbl_time { 
                    #[watch]
                    set_label: model.time_str
                }
```

Is it possible to access the `model` *(and also `sender` would be great)* from **WidgetTemplates**?

AaronErhardt:
Widget templates are so far just a simple abstraction for grouping widgets together. This makes them easy to use because they don't have any state and don't require any knowledge about component they are used in.

To make it possible to access the model in widget templates, you'd need to specify the type of the component for each widget template, which makes them less useful as templates because then they only work for a single component. It would be possible to introduce a `Update` type to the `WidgetTemplate` trait to allow the template to access this data, but I don't know whether templates are really the best abstraction for this, or whether we should look for other solutions...

PR #526, opened 2023-09-08
core: WIP factory-builder

AaronErhardt:
@chriskilding I've had some time to work on a solution for the factory builder implementation. You might rebase your PR on this branch as it does most of the heavy lifting, but is still a bit rough around the edges. If you have some time to polish the new interfaces and update the examples, that'd be awesome.


chriskilding:
Thanks! I've now incorporated these changes into my PR.

We should keep this PR around while we get the other one over the line, after that we can close this one.

PR #531, opened 2023-09-12
macros: Allow trailing commas in the last match arm

### LINKED ISSUE #506: Trailing commas are not supported
Following code doesn't compile, but after just removing trailing comma from gtk::Label block it works fine.
```
use relm4::{ComponentParts, ComponentSender, gtk, RelmApp, SimpleComponent};

struct App {
    value: Value,
}

enum Value {
    Single,
}

#[relm4::component(pub)]
impl SimpleComponent for App {
    type Init = ();
    type Input = ();
    type Output = ();

    view! {
        gtk::Window {
            match model.value {
                Value::Single => {
                    gtk::Label {
                        set_label: "Value is higher than 2",
                    },
                  // ^ unexpected token
                }
            },
        }
    }

    fn init(
        _init: Self::Init,
        _root: &Self::Root,
        _sender: ComponentSender<Self>,
    ) -> ComponentParts<Self> {
        let model = App {
            value: Value::Single,
        };

        let widgets = view_output!();

        ComponentParts {
            model,
            widgets,
        }
    }
}

fn main() {
    let app = RelmApp::new("relm4.example.trailing_coma");
    app.run::<App>(());
}

```

AaronErhardt:
#### Summary

Fixes #506

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
 ### Fixed
 
 + macros: Fix track attribute for template children
++ macros: Allow trailing commas in the last match arm
 
 ## 0.7.0-alpha.1 - 2023-8-28
 
--- a/examples/macro_reference.rs
+++ b/examples/macro_reference.rs
@@ -90,7 +90,7 @@ impl SimpleComponent for App {
                         gtk::Label {
                             set_label: "Value is higher than 2",
                         }
-                    }
+                    },
                 },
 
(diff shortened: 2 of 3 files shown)


Skaldebane:
Thanks, that was quick!

PR #533, opened 2023-09-15
macros: Improve codegen to prevent incorrect lints

### LINKED ISSUE #532: `relm4_macros::view!` triggers `clippy::used_underscore_binding` lint.
# My use case

I'm using the `relm4_macros::view` macro standalone for convenience when working with `gtk4` or `libadwaita`. I noticed that clippy generates lints on pretty much every `view!` invocation. This is probably due to [`Span::call_site()`](https://docs.rs/proc-macro2/latest/proc_macro2/struct.Span.html#method.call_site) invocations that should really be [`Span::mixed_site()`](https://docs.rs/proc-macro2/latest/proc_macro2/struct.Span.html#method.mixed_site).

# Example

```
use relm4_macros::view;
use gtk4 as gtk;

fn main() {
  view! {
    application = gtk::Application {
      set_application_id: ("some.application.id"),
      connect_activate: build_ui,
    },
}

fn build_ui(app: &gtk::Application) {
    main_window = gtk::ApplicationWindow {
      set_application: Some(application),

      // this label here is the culprit
      #[wrap(Some)]
      set_content = &gtk::Label {
        set_text: "test",
      },
    }
  }
}
```

The unnamed `gtk::Label` here will get created under a name like `_gtk_label_0`.
This name has a `call_site` span and is thus recognized by clippy.

# Workaround

Add `#[allow(clippy::used_underscore_binding)]` at the macro call sites or at the crate root.

# Resolution

I haven't quite determined the code in `relm4-macros/widgets` that is responsible for giving the generated identifiers their scope, but just from looking at it at first glance, most `Ident`s are created using `Span::call_site`, even if the name has been generated internally. For good macro hygiene the generated ones should probably use `mixed_site` hygiene. At least the `_[PATH]_[NUMBER]` style identifiers should really have `mixed_site` hygiene.

Changing the public symbols generated by the `view!` macro would by a subtle but breaking change.
Annotating all generated `let` expressions with `#[allow(clippy::used_underscore_binding)]` would at least fix the strange lint the user can't do anything about.

# Notes

I want to add that I really love the approach taken by the `view!` macro. It helps readability and reduces boilerplate to a degree that makes the lack of automatic formatting pretty irrelevant. Thank you so much for imagining a better GTK development experience!

AaronErhardt:
#### Summary

Fixes #532

@PJungkamp it seems like `mixed_site` would have been more correct in some cases, but it didn't cause the problem. The actual cause was that the span of the type (like `gtk::Box`) was used for the automatically generated ident (like `_gtk_box_1`), so it looked to clippy like it was added by the user. This behavior was initially added because I thought it might help with error reporting, but in the end it can only cause problems if the macro does something wrong.

Can you test if this fixes the problem for you?

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -55,6 +55,7 @@
 
 ### Fixed
 
++ macros: Improve spans in code generation to prevent false positive lints
 + macros: Fix track attribute for template children
 + macros: Allow trailing commas in the last match arm
 
--- a/examples/macro_reference.rs
+++ b/examples/macro_reference.rs
@@ -1,3 +1,14 @@
+// Catch potential problems with clippy and the macro
+#![warn(
+    missing_debug_implementations,
+    rust_2018_idioms,
+    unreachable_pub,
+    unused_qualifications,
+    clippy::cargo,
+    clippy::must_use_candidate,
+    clippy::used_underscore_binding
+)]
+
 use gtk::prelude::{
     BoxExt, ButtonExt, GridExt, GtkWindowExt, ObjectExt, OrientableExt, ToggleButtonExt, WidgetExt,
 };
--- a/relm4-macros/src/menu/gen.rs
+++ b/relm4-macros/src/menu/gen.rs
@@ -72,7 +72,7 @@ impl MenuItem {
 
 impl SubMenu {
     fn submenu_stream(&self, parent_ident: &Ident) -> TokenStream2 {
-        let name = Ident::new(&format!("_{parent_ident}"), Span2::call_site());
+        let name = Ident::new(&format!("_{parent_ident}"), Span2::mixed_site());
         let gtk_import = crate::gtk_import();
         let expr = &self.expr;
 
--- a/relm4-macros/src/menu/parse.rs
+++ b/relm4-macros/src/menu/parse.rs
@@ -99,5 +99,5 @@ fn section_name() -> Ident {
 
     let value = COUNTER.fetch_add(1, Ordering::Relaxed);
 
-    Ident::new(&format!("_section_{value}"), Span2::call_site())
+    Ident::new(&format!("_section_{value}"), Span2::mixed_site())
 }
--- a/relm4-macros/src/util.rs
+++ b/relm4-macros/src/util.rs
@@ -56,7 +56,7 @@ pub(super) fn strings_to_path(strings: &[&str]) -> Path {
         .iter()
         .map(|string| -> PathSegment {
             PathSegment {
-                ident: Ident::new(string, Span2::call_site()),
+                ident: Ident::new(string, Span2::mixed_site()),
                 arguments: PathArguments::None,
             }
         })
@@ -98,7 +98,7 @@ pub(super) fn verbatim_impl_item_fn(
             unsafety: None,
             abi: None,
             fn_token: syn::token::Fn::default(),
-            ident: Ident::new(name, Span2::call_site()),
+            ident: Ident::new(name, Span2::mixed_site()),
             generics: syn::Generics {
                 lt_token: None,
                 params: Punctuated::default(),
--- a/relm4-macros/src/view.rs
+++ b/relm4-macros/src/view.rs
@@ -17,7 +17,7 @@ pub(super) fn generate_tokens(input: TokenStream) -> TokenStream {
     } = view_widgets.generate_streams(
         &TraitImplDetails {
             vis: None,
-            model_name: Ident::new("_", Span2::call_site()),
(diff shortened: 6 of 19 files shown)


PJungkamp:
You're right, this branch fixes the lint! Though I still don't like the fact that something like this works:

```rs
        view! {
            main_window = adw::ApplicationWindow {
                set_application: Some(application),

                #[wrap(Some)]
                set_content = &adw::NavigationSplitView {
                    set_sidebar: Some(&adw::NavigationPage::new(&sidebar_page, "Sidebar")),
                    set_content: Some(&adw::NavigationPage::new(&main_page, "Content")),
                },
            },
        };
        print("{_adw_navigationsplitview_0:?}"); // referring to an identifier that I never wanted to be in scope.
```

This is roughly the case in my code that made me notice the lint.

PJungkamp:
I suspect that all invocations of `idents_to_snake_case` are creating names in the calling scope that are nameable but shouldn't be. The `__returned_widget` and `_conditional_widget` names are still affected.

There are still at least 2 cases not covered:

- For `_conditional_widget_[number]`
  https://github.com/Relm4/Relm4/blob/273e167b1751ca122c3846ed8db6c8b19f901765/relm4-macros/src/widgets/parse/conditional_widget.rs#L55-L58
- For `__returned_widget_[number]`
  https://github.com/Relm4/Relm4/blob/273e167b1751ca122c3846ed8db6c8b19f901765/relm4-macros/src/widgets/parse/returned_widget.rs#L33-L36

Both types of widgets are nameable outside the `view!` macro.

I also found it very surprising that the `idents_to_snake_case` function is more of a `create_unique_ident`, that adds an underscore to the front and creates a unique suffix number on every invocation. Would you consider renaming the function and removing the `span` argument alltogether, so a generated ident always has mixed site span?

PJungkamp:
Check out the changes on my `span-fix` branch here: https://github.com/PJungkamp/Relm4/compare/nix-direnv...PJungkamp:Relm4:span-fix

Can you just pull the changes into your branch? (I've also got my Nix dev env committed, don't pull that :D)

AaronErhardt:
I didn't have time this weekend but I'll review your commits tomorrow. You can also just open a PR and target this branch btw.

AaronErhardt:
The patches look good! With the large amount of test coverage in the Relm4 repo I'd be confident to merge this if the CI passes on the changes.

If you have some time, it would be great if you could separate those three commits from your nix environment (with cherry-pick, for example) and open a PR. I could do it as well, but it would require me a couple of extra steps to stitch everything together.

PJungkamp:
Done! I think that's about all I'd want to put into this PR.

I just moved to a copied the branch and then did an interactive rebase onto your `span-fix` dropping the unrelated commits.
Did you check the detailed meaning of mixed site hygiene? Since it only applies to identifiers the `*Widgets` type and the `__DoNotReturnManually` hack is still unhygienic.

The `*Widgets` types should probably stay nameable. It's generation is also documented in relm4 book, so I think it's fine.
I find the `__DoNotReturnManually` hack strange. I was really suprised that the `pre_view` function, isn't a function. It's really weird to take the function body and just emplace it in another scope. I also didn't understand the justification for disallowing an early return.

Greetings from Aachen, seems like we're working in similar fields. :D

AaronErhardt:
> I find the __DoNotReturnManually hack strange

I know, it looks strange. The thing is though, if you return from `pre_view` (which you usually could), the macro-generated update code wouldn't run anymore, which has once confused someone in the past.

The more idiomatic approach would probably to use something similar to `init()` where you can write manual code around the `view_output!()` macro (in the `view_update()` method) to know where the codegen is pasted.

In general, I think the macro needs a complete rework, but I don't want to do it with syn because it is often quite ugly to work with and makes it hard to implement correct, predictable and scalable macros.

> Greetings from Aachen, seems like we're working in similar fields. :D

Thanks, greetings from Augsburg! It's nice to see that Ferris has made it to Aachen as well :D
I recently finished my bachelor and will start with my master studies in October btw. :)

Mir ist gerade aufgefallen, wir könnten uns auch einfach gleich auf Deutch schreiben 😆

AaronErhardt:
Thanks for your help with this @PJungkamp!

PJungkamp:
Ich benutze das das `view!` Macro immer mal wieder um gtk-rs style code I verschönern. (Manchmal auch mit anderen Crates)

Gestern bin ich dann wieder über meinen offenen Relm4 PR gestolpert.

Danke für die super Arbeit an Relm4!

PR #534, opened 2023-09-17
Add a `SimpleComboRow` helper for `adw::ComboRow`

sgued:
#### Summary

This PR adds:

- A `SimpleComboRow` component, as an easy to use wrapper around `adw::ComboRow`, similar to the existing `SimpleComboBox`
- Examples for `SimpleComboBox` and `SimpleComboRow`

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

The examples are a bit raw, I can make them a bit prettier if necessary.

### DIFF
--- a/relm4-components/Cargo.toml
+++ b/relm4-components/Cargo.toml
@@ -29,9 +29,10 @@ reqwest = { version = "0.11.18", optional = true }
 tracker = "0.2.1"
 
 [features]
-default = []
+default = ["libadwaita"]
 dox = ["web"]
 web = ["reqwest"]
+libadwaita = ["relm4/libadwaita"]
 
 [[example]]
 name = "web_image"
--- a/relm4-components/Cargo.toml
+++ b/relm4-components/Cargo.toml
@@ -37,3 +37,7 @@ libadwaita = ["relm4/libadwaita"]
 [[example]]
 name = "web_image"
 required-features = ["web"]
+
+[[example]]
+name = "adw_combo_row"
+required-features = ["libadwaita"]
--- a/relm4-components/examples/adw_combo_row.rs
+++ b/relm4-components/examples/adw_combo_row.rs
@@ -1,0 +1,87 @@
+use adw::prelude::*;
+use relm4::{
+    adw, gtk, Component, ComponentController, ComponentParts, ComponentSender, Controller, RelmApp,
+    SimpleComponent,
+};
+use relm4_components::simple_adw_combo_row::SimpleComboRow;
+
+#[derive(Debug)]
+enum AppMsg {
+    Selected(usize),
+}
+
+struct App {
+    combo_row: Controller<SimpleComboRow<&'static str>>,
+    selected_variant: usize,
+}
+
+#[relm4::component]
+impl SimpleComponent for App {
+    type Init = ();
+    type Input = AppMsg;
+    type Output = ();
+
+    view! {
+        #[name = "app"]
+        adw::Window {
+            set_default_size: (300, 100),
+            set_title: Some("Libadwaita SimpleComboRow"),
+
+            gtk::Box {
+                set_orientation: gtk::Orientation::Vertical,
+                set_vexpand: true,
+
+                #[name = "sidebar_header"]
+                adw::HeaderBar {
+                    #[wrap(Some)]
+                    set_title_widget = &adw::WindowTitle {
+                        set_title: "Libadwaita SimpleComboRow",
+                    },
+                    set_show_end_title_buttons: false,
+                },
+
+                adw::PreferencesGroup {
+                    #[local_ref]
+                    combo_row -> adw::ComboRow,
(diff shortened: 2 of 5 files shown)


sgued:
The build errors seem unrelated.

relm4-components/Cargo.toml
```diff
@@ -27,12 +27,18 @@ once_cell = "1.18"
...
 reqwest = { version = "0.11.18", optional = true }
 tracker = "0.2.1"
+adw = { version = "0.5", optional = true, package = "libadwaita" }
 
 [features]
-default = []
+default = ["libadwaita"]
 dox = ["web"]
 web = ["reqwest"]
+libadwaita = ["adw", "relm4/libadwaita"]
```
AaronErhardt, line 36: `libadwaita = ["relm4/libadwaita"]` is enough, you can import adwaita through `relm4::adw`.

relm4-components/examples/adw_combo_row.rs
AaronErhardt, line 1: Examples for libadwaita have so far been placed in [`examples/libadwaita`](https://github.com/Relm4/Relm4/tree/main/examples/libadwaita). The example will compile as expected in the regular examples folder because "libadwaita" is enabled by default currently. But once you disable default features, it doesn't work anymore.

I think I should probably add a CI check for that at some point 🤔

relm4-components/src/simple_adw_combo_row.rs
```diff
@@ -0,0 +1,121 @@
...
+
+impl<E> SimpleComboRow<E>
+where
+    E: ToString,
+{
+    fn render(&self, combo_box: &adw::ComboRow) {
+        use std::ops::Deref;
+        if combo_box.model().is_some() {
+            unreachable!()
+        }
```
AaronErhardt, line 105: This isn't exactly unreachable. It is possible for the user to add entries by accessing the combo row as root widget manually. And I think the `UpdateData` message would also trigger a panic here.
The question is whether this needs to be checked at all. If the user does stuff manually, they should know what they do and expect unexpected behavior.

relm4-components/src/simple_adw_combo_row.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        let variants: Vec<String> = self.variants.iter().map(ToString::to_string).collect();
+        let variants: Vec<&str> = variants.iter().map(String::deref).collect();
+        combo_box.set_model(Some(&StringList::new(&variants)));
```
AaronErhardt, line 109: You should be able to directly collect into the `StringList` in just one line: https://gtk-rs.org/gtk4-rs/git/docs/gtk4/struct.StringList.html#impl-FromIterator%3CString%3E-for-StringList

relm4-components/src/simple_adw_combo_row.rs
sgued: Doesn't the `required-features` field make this work for `--all-targets`?

relm4-components/examples/adw_combo_row.rs
sgued: Doesn't the `required-features` field make this work for `--all-targets`?

relm4-components/src/simple_adw_combo_row.rs
sgued: Good catch!

I removed the check.

relm4-components/examples/adw_combo_row.rs
sgued: I'm not sure I understand exactly what you want me to do with this example.

I initially put it there because there didn't appear to be any example using `relm4-components` in `/examples`.

I can create a separate folder in `/relm4-components/examples/libadwaita` or put it in the top-level `/examples/libadwaita` , whatever you prefer.

AaronErhardt: Sorry, my bad. I didn't read the whole path and thought is was going to be added to `examples`, but it's actually `relm4-components/examples`. You can simply ignore my previous comments about this.

relm4-components/src/simple_adw_combo_row.rs
sgued: Oh, that's much simpler and performant thanks.

It's crazy how far the GTK actually go!

relm4-components/Cargo.toml
sgued: Right!

Fixed it.

sgued:
Thanks for the review. I fixed the comments and push the changes. Following https://github.com/Relm4/Relm4/pull/535#issuecomment-1724221569 I removed the `SimpleComboBox` example.

Will you first merge this or #535 ? There will be a conflict with the `CHANGES.md` file.

AaronErhardt (approved):
Thanks a lot! Let's just merge this, I can take care of the conflicts myself.

PR #535, opened 2023-09-17
Add an example how to use SimpleComboBox

marmistrz:
#### Summary
I found it not so easy to start using `SimpleComboBox` in my project (the book doesn't seem to explain much about injecting components into the view, the only examples I found were about dialogs, i.e., `#[local_ref]` is not really explained there).

So I created an example how to use it.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy (unrelated lints in `bindings.rs`)
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,8 @@
 ### Added
 
 + core: Add `AsyncComponentStream` to supports streams for async components as well
-
++ components: Add an example how to use `SimpleComboBox`
+ 
 ### Changed
 
 + core: Just require `AsRef<gtk::Window>` in `RelmApp::run`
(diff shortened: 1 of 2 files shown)


sgued:
Note that this is implemented in #534

marmistrz:
Submitted 3 hours before! I'm fine with using either example, then.

AaronErhardt:
First of all, thank you both @marmistrz and @sgued!

It's crazy, the timing couldn't have been worse. For months there was barely any example added and now there are two at the same time about the same feature.

To resolve the conflicting implementations, I suggest that we use this version of the combo box example for pure GTK and use (only) the libadwaita example of your version @sgued. Technically, the examples are both great. Yet, the GTK and libadwaita examples in #534 are almost identical, so I think using an example from another author simply adds just a little bit more value. Also, I like the idea of the language selector, which is a pretty realistic use-case.

AaronErhardt (approved):
Thanks!

PR #536, opened 2023-09-18
Improve macro hygiene

PJungkamp:
See #533 for more information.

### DIFF
--- a/relm4-macros/src/widgets/gen/assign/assign_property.rs
+++ b/relm4-macros/src/widgets/gen/assign/assign_property.rs
@@ -100,31 +100,30 @@ impl AssignProperty {
                     }
                 }
                 (true, false) => {
-                    quote_spanned! {
-                        span => if let Some(__p_assign) = #assign {
+                    quote! {
+                        if let Some(assign) = #assign {
                             #block_stream
-                            #assign_fn(#self_assign_args __p_assign #args) #chain;
+                            #assign_fn(#self_assign_args assign #args) #chain;
                             #unblock_stream
                         }
                     }
                 }
                 (false, true) => {
-                    quote_spanned! {
-                        span =>
-                            #block_stream
-                            for __elem in #assign {
-                                #assign_fn(#self_assign_args __elem #args) #chain;
-                            }
-                            #unblock_stream
+                    quote! {
+                        #block_stream
+                        for elem in #assign {
+                            #assign_fn(#self_assign_args elem #args) #chain;
+                        }
+                        #unblock_stream
                     }
                 }
                 (true, true) => {
                     quote_spanned! {
                         span =>
                             #block_stream
-                            for __elem in #assign {
-                                if let Some(__p_assign) = __elem {
-                                    #assign_fn(#self_assign_args __p_assign #args) #chain;
+                            for elem in #assign {
+                                if let Some(assign) = elem {
+                                    #assign_fn(#self_assign_args assign #args) #chain;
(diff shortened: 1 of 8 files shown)


relm4-macros/src/widgets/gen/assign/assign_property.rs
```diff
@@ -100,31 +100,30 @@ impl AssignProperty {
                     }
```
AaronErhardt, line 100: Why keep the `quote_spanned` in line 96? Is there something special there?

relm4/src/binding/bindings.rs
```diff
@@ -41,6 +41,7 @@ macro_rules! binding {
         }
 
         #[allow(missing_docs)]
+        #[allow(clippy::must_use_candidate)]
```
AaronErhardt, line 44: This is a regression in `glib-macros` and I've just sent a PR to fix this. Ironically, the have the same problem: they recently changed `quote` to `quote_spanned` (with nonsensical spans) and now clippy is throwing a warning that wasn't there before.

Which means that this should not be fixed from our side (even if it breaks the CI temporarily).

PJungkamp: `quote_spanned` does make sense there, as problems regarding the function are now properly attributed to the `#[property]` helper attribute instead of the `Properties` identifier in the `#[derive]` attribute. Using `quote_spanned` is generally good practice in more complex macros. The problem with `quote_spanned` in Relm4 are the double underscore identifiers created in the `quote_spanned` invocation.

relm4-macros/src/widgets/gen/assign/assign_property.rs
PJungkamp: That `quote_spanned` does not introduce an identifier into scope. `quote_spanned` invocations are fine as long as there are no new identifiers created inside.


This is problematic, as `__return` has a call site span.
```rust
quote_spanned! {Span2::call_site()=>
    let __name = #expr;
}
```

This isn't problematic as `quote_spanned` does not touch the span of substitutions:
```rust
quote_spanned! {Span2::call_site()=>
    let #name = #expr;
}
```

Instead of replacing `quote_spanned` with `quote`, I could have also just made the offending identifier a substitution with proper span.
```rust
let name = Ident::new("name", Span2::mixed_site());
quote_spanned! {Span2::call_site()=>
    let #name = #expr;
}

File: relm4/src/binding/bindings.rs (hunk 2 above)
Comment (AaronErhardt) [reply relm4/src/binding/bindings.rs:44]: Yeah, the actual problem in Relm4 is that I've started with `quote` which resulted in poor error reporting and then thought that `quote_spanned` would be always better, which was sometimes true and sometimes not. And then I just kept it as it was.

It's not easy to do generate properly spanned code with quote in general though, because if I think about it, you'd probably need to split some `quote` invocations into several `quote` and `quote_spanned` invocations and merge their output into one stream to get it right in some places...

Comment (AaronErhardt) [reply relm4/src/binding/bindings.rs:44]: Anyway, `#[allow(clippy::must_use_candidate)]` still needs to be removed before merging, because it's not a problem caused by Relm4 and was fixed in glib-macros (just not published on crates.io yet).

Comment (PJungkamp) [reply relm4/src/binding/bindings.rs:44]:
```suggestion
```

I'll remove it. Do you have any further concerns regarding this PR?

AaronErhardt (approved):
Looks good. Thanks for the PR!

PR #538, opened 2023-09-24
Return Result from `FactorySender#output` method

chriskilding:
#### Summary

Return the `Result` object from `FactorySender#output`, instead of silently unwrapping it. This allows app developers to handle the `Result` as they wish (rather than it crashing the app if no handler function is attached).

This brings the signature of this `output` method into line with the `ComponentSender` and `AsyncComponentSender`.

#### Checklist

- [ ] cargo fmt
- [ ] cargo clippy
- [ ] cargo test
- [ ] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Changed
+
++ core: Return `Result` from `FactorySender#output` method (so that errors are not silently unwrapped)
+
 ## 0.7.0-beta.1 - 2023-9-23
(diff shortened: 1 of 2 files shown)


AaronErhardt:
Thanks!

Btw. updating the Cangelog would really help. I almost forgot to add the factory changes to the beta release and having them in the PR itself makes it much easier to handle :)

AaronErhardt (approved):
Thanks!

PR #539, opened 2023-09-24
Run cargo clippy check in a parallel CI job

chriskilding:
#### Summary

Parallelize the `cargo clippy` check by running it in its own GH Actions job.

This will speed up the build, and also allow `cargo test` to execute even if Clippy fails, allowing us to see any test errors in addition to Clippy errors.

#### Checklist

- [ ] cargo fmt
- [ ] cargo clippy
- [ ] cargo test
- [ ] updated CHANGES.md

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -27,6 +27,29 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust }}
+
+      - name: "test"
+        run: xvfb-run --auto-servernum cargo test
+
+      - uses: bcomnes/cleanup-xvfb@v1.0.9
+
+  clippy:
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/relm4/relm4/docs:latest
+    strategy:
+      matrix:
+        rust:
+          - stable
+          - nightly
+          - "1.70.0"
+    env:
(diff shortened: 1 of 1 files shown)


chriskilding:
One additional question on this: is the Rust build matrix of stable/nightly/1.70 also required for `cargo test`, or is it only needed for `cargo clippy`?

AaronErhardt:
> One additional question on this: is the Rust build matrix of stable/nightly/1.70 also required for cargo test, or is it only needed for cargo clippy?

Running tests for all toolchains could help catching problems early, but on the other side, it is very unlikely that this will happen. If nighly clippy does not complain, yet still a test case fails on nightly which passes on stable, that would be a pretty big regression in the compiler and Relm4 would not be the first crate to notice it. TL;DR it should be reasonable to only run tests on stable. 

Another benefit of only testing on stable would be that we could remove [this line](https://github.com/Relm4/Relm4/blob/main/relm4-macros/src/lib.rs#L639) and probably the entire `rustversion` crate as dependency.

chriskilding:
Great - have updated the PR with that change

chriskilding:
Is there anything more to do on this PR before it's merged?

AaronErhardt:
Thanks!

> Is there anything more to do on this PR before it's merged?

No, I've simply been quite busy, otherwise I'd have merged it earlier.

PR #541, opened 2023-09-26
Hotfix: make factory and grid_factory examples work again

chriskilding:
#### Summary

I noticed a couple of the examples stopped compiling with the latest change in 828c940. This hotfix should make the build work again.

#### Checklist

- [ ] cargo fmt
- [ ] cargo clippy
- [ ] cargo test
- [ ] updated CHANGES.md

### DIFF
--- a/examples/factory.rs
+++ b/examples/factory.rs
@@ -57,7 +57,7 @@ impl FactoryComponent for Counter {
             gtk::Button {
                 set_label: "Up",
                 connect_clicked[sender, index] => move |_| {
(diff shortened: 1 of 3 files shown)


chriskilding:
This is a minor PR as they go, but I'll add a changelog entry if you think it merits one

AaronErhardt (approved):
Thanks!

Btw. you only need to update the changelog if you add something meaningful or change/fix the actual relm4 crate. In this case, it's better to not mention anything because it doesn't affect our users.

PR #543, opened 2023-09-27
Show how to access variants when the combo box is already launched

### LINKED ISSUE #540: [Ergonomics/performance] Can we avoid storing the SimpleComboBox variants twice
Currently only the index is returned, but this requires the variants to be stored _twice_: once in the `SimpleComboBox` and once in the closure. (both `launch` and `forward` consume the component, so it's impossible to use it in the `forward` closure).

However, returning `E` would require `E: Clone` and returning `&E` gets us into lifetime troubles.

marmistrz:
Closes #540

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [ ] updated CHANGES.md (no need to)

### DIFF
--- a/relm4-components/examples/combo_box.rs
+++ b/relm4-components/examples/combo_box.rs
@@ -7,9 +7,6 @@ use relm4_components::simple_combo_box::SimpleComboBox;
 
 type ComboContent = &'static str;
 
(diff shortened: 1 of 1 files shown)


AaronErhardt (approved):
Looks good, thanks!

PR #546, opened 2023-10-02
Implement traits to make working with the new adwaita NavigationView easier

myuujiku:
#### Summary

Implement Relm traits for `NavigationView` and `NavigationPage` introduced in libadwaita v1.4.
(I did not touch `NavigationSplitView`)

This doesn't close any issue, as far as I know, but makes working with the new widgets easier. I've almost opened an issue but decided to try to fix the problem myself instead. This is my first time working with this code base so please let me know if there are any issues. I tried out the changes in a small test application and didn't encounter any issues.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
++ core: Implement `RelmContainerExt` for libadwaita 1.4's `NavigationView`
++ core: Implement `RelmSetChildExt` for libadwaita 1.4's `NavigationPage`
+
 ### Changed
 
 + core: Return `Result` from `FactorySender#output` method (so that errors are not silently unwrapped)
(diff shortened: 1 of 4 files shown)


AaronErhardt (changes requested):
Thanks for your contribution! (And sorry for my slow response)

The code looks good overall, but I think we need `#[cfg(all(feature = "libadwaita", feature = "gnome_45")]` for all `cfg` attributes because theoretically, you can use Relm4 with recent GTK versions still without libadwaita.

CHANGES.md
```diff
@@ -12,6 +12,8 @@
 
 + core: Add `AsyncComponentStream` to supports streams for async components as well
 + core: Builder pattern for initializing factories similar to regular components
++ core: Implement `RelmContainerExt` for libadwaita 1.4's `NavigationView`
++ core: Implement `RelmSetChildExt` for libadwaita 1.4's `NavigationPage`
```
AaronErhardt, line 16: And I just noticed, this has to moved to `## Unreleased` because we can't add the current changes to an already released version of Relm4. So please add a `### Added` category underneath `## Unreleased` :)

myuujiku: Ah, my bad! I'll quickly fix those things.

myuujiku: I hope this does it. If there are any other issues let me know! ^^

And don't worry about the response time. I'm aware that you are busy and that is completely fine.

AaronErhardt (approved):
Looks good, thanks!

PR #553, opened 2023-10-11
core: Add builder pattern for async factories

AaronErhardt:
#### Summary

[See changelog](https://github.com/Relm4/Relm4/compare/async-factory-builder?expand=1#diff-d975bf659606195d2165918f93e1cf680ef68ea3c9cab994f033705fea8238b2) + some minor cleanups

I think using `::builder().launch(arg)` is more natural then `::builder(arg).launch()`.

@chriskilding if you have time, can you have a look over the changes?

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,9 +2,19 @@
 
 ## Unreleased
 
+### Added
+
++ core: Add builder for initializing async factories
++ core: Add `launch_default()` method to factory builders for launching with default parent widget
+
 ### Changed
 
 + core: Return `Result` from `FactorySender#output` method (so that errors are not silently unwrapped)
++ core: Move parameters from `builder()` to `launch()` in factories for more consistency
+
+### Fixed
+
++ core: Properly re-export factory builder types and document them
 
 ## 0.7.0-beta.1 - 2023-9-23
 
--- a/examples/entry.rs
+++ b/examples/entry.rs
@@ -32,7 +32,7 @@ impl FactoryComponent for Counter {
             #[watch]
             set_label: &self.value.to_string(),
             connect_clicked[index] => move |_| {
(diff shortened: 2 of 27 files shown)


chriskilding:
One thing I found looking at the examples: with the output() method now returning the Result (rather than hiding it), if you do `sender.output(...).unwrap()` in the component's event handler *and* you use `.detach()` to start the component, the app will crash whenever that event handler runs. The reason being that it's detached... so there's nothing listening on the other end of the output channel.

As a result it would be better to do `let _ = sender.output(...)` in the examples; this lets it keep going if the component is detached.

chriskilding:
As for whether to put extra arguments into `builder()` or `launch()`, the only constraints I used for writing the sync factory builder were

- follow the conventions laid down by the regular component builder, to maintain consistency
-  if the extra arguments should influence the hash value of the factory object or the factory builder object, then this might change where you inject them

I do think keeping consistency with the regular component builder is very important for the developer experience, so `launch_default()` from this perspective is odd; there's no method named like this in the regular component builder.

AaronErhardt:
I'll merge those changes for now because I think they look reasonable and will wait for more feedback in the next beta release.

chriskilding:
Some feedback after playing with this in an example project...

It works well. Having control over the returned `Result` is helpful (as I've mentioned). The `launch_default()` method is nice in that I no longer have to instantiate and pass the parent element to the method. 

One small improvement; most of the time I find I just want to launch with the implicit default parent object. I'd therefore prefer if the method for this was called `launch()`, while the method to launch with a custom parent element object was called the longer version of the name. For example it could be called `launch_with(element)` or `launch_with_parent(element)`.

PR #559, opened 2023-10-17
core: Fix overriding the visibility of the main window

### LINKED ISSUE #558: Unable to hide an application window
I have a use case for a file synchronization program I'm working on (see [hwittenborn/celeste](/hwittenborn/celeste)). In that program, there's a main window that shows up while syncing files, but if a user hasn't logged in at all yet, a login window gets displayed instead.

I'd like to implement this by having an `AppModel` and `LoginModel` struct. When the user still needs to log in, I'd like the window for `AppModel` to be hidden, but I can't figure out how I'd do it in Relm4.

Here's a test case that demonstrates the problem:

```rust
use relm4::prelude::*;
use adw::prelude::*;

struct AppModel;

#[relm4::component]
impl SimpleComponent for AppModel {
    type Input = ();
    type Output = ();
    type Init = ();

    view! {
        adw::ApplicationWindow {
            set_visible: false,

            adw::HeaderBar {}
        }
    }

    fn init(
        init: Self::Init,
        root: &Self::Root,
        sender: ComponentSender<Self>,
    ) -> ComponentParts<Self> {
        let model = Self;
        let widgets = view_output!();
        ComponentParts { model, widgets }
    }
}

fn main() {
    let app = RelmApp::new("com.example.Testing");
    app.run::<AppModel>(());
}
```

I also made an example using libadwaita directly to make sure I wasn't just calling an incorrect function or something, but it indeed works fine:

```rust
use adw::{Application, ApplicationWindow, prelude::*, gio, glib::{self, MainContext, Priority}};
use std::{time::Duration, thread};

fn main() {
    let app = Application::builder()
        .application_id("com.hunterwittenborn.Testing")
        .build();
    app.connect_activate(|app| {
        let window = ApplicationWindow::builder()
            .application(app)
            .title("Hello, World!")
            .build();
        window.present();

        // Close the window after waiting two seconds.
        let (sender, receiver) = MainContext::channel(Priority::default());
        gio::spawn_blocking(move || {
            thread::sleep(Duration::from_secs(2));
            sender.send(()).unwrap();
        });
        receiver.attach(None, move |_| {
            window.set_visible(false);
            glib::ControlFlow::Continue
        });
    });
    app.run();
}
```

Is there any way to get what I'm wanting to work with Relm4?

AaronErhardt:
#### Summary

Fixes #558

See changlog

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,8 +2,19 @@
 
 ## Unreleased
 
-+ core: Move internal initialization to `gtk::Application::startup` signal handler
+### Added
+
++ core: Make `into_stream()` method on `Receiver` public
 + examples: Add libadwaita `Toast` example
+
+### Changed
+
++ core: Move internal initialization to `gtk::Application::startup` signal handler
++ core: Remove deprecated RelmApp methods
+
+### Fixed
+
++ core: Setting the visibility of the main window isn't overridden by `RelmApp` anymore
 
 ## 0.7.0-beta.2 - 2023-10-14
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 ### Added
 
++ core: Add `visible_on_activate()` to `RelmApp` to prevent the app window from being visible immediately
 + core: Make `into_stream()` method on `Receiver` public
 + examples: Add libadwaita `Toast` example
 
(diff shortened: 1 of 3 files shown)


hwittenborn:
Hey @AaronErhardt, is there anything stopping this from being merged? There's some contributions I'm looking into making to Relm4, but I'm trying to avoid having a bunch of stacked PRs.

AaronErhardt:
Sorry, I forgot about this PR, but it should be save to merge.

PR #564, opened 2023-10-23
core: Pass owned root to Component::init

### LINKED ISSUE #563: Why does `SimpleComponent` take in `&Self::Root`, but `SimpleAsyncComponent` takes in `Self::Root`?
### Discussed in https://github.com/orgs/Relm4/discussions/560

<div type='discussions-op-text'>

<sup>Originally posted by **hwittenborn** October 17, 2023</sup>
I was migrating some components from using `SimpleComponent` to using `SimpleAsyncComponent`, and was greeted with this pretty confusing compiler error (at least until I found the solution):

```rust
error[E0195]: lifetime parameters or bounds on method `init` do not match the trait declaration
  --> src/launch.rs:37:14
   |
37 |       async fn init(
   |  ______________^
38 | |         init: Self::Init,
39 | |         root: &Self::Root,
40 | |         sender: AsyncComponentSender<Self>,
41 | |     ) -> AsyncComponentParts<Self> {
   | |_____^ lifetimes do not match method in trait
```

Changing `root: &Self::Root` to `root: Self::Root` fixed the issue, but this wasn't apparent until I manually checked the trait definitions for both.

[The page in the book talking about async components](https://relm4.org/book/stable/threads_and_async/async.html) made me believe that I wouldn't have to think about it:

> Asynchronous components and factories are almost identical compared to regular components and factories.
> The only major difference is that they have asynchronous `init`, `update` and `update_cmd` methods.
> This allows you to `await` almost everywhere from within the component.

If there's some internal reason that `root` requires `Self::Root` instead of `&Self::Root` I think it'd be important to note, as a spent a hot minute trying to figure out what went wrong while swapping out the trait being used.</div>

AaronErhardt:
#### Summary

Fixes #563

I'd like some feedback first. This change is small, but affects existing code bases in a lot of places.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -53,7 +53,7 @@ jobs:
         run: cargo clippy --features "all" -- --deny warnings
 
       - name: "clippy check examples"
-        run: cargo clippy --examples -- --deny warnings
+        run: cargo clippy --examples --all-features -- --deny warnings
 
       - name: "clippy check libadwaita examples"
         run: cd examples/libadwaita && cargo clippy --examples -- --deny warnings
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
 ### Changed
 
 + core: Make `set_global_css` and `set_global_css_from_file` methods of `RelmApp` to prevent calling them before initializing GTK 
++ core: Always pass the `Root` as owned parameter (without reference) (affects `Component::init`, `Factory::init_widgets`, `AsyncComponent::init_loading_widgets`, `AsyncFactory::init_widgets` and `AsyncFactory::init_loading_widgets`)
 + core: Move internal initialization to `gtk::Application::startup` signal handler
 + core: Remove deprecated RelmApp methods
 + components: Increase flexibility of `Alert` component
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ### Added
 
 + core: Add resizeable/expandable column functionality to `RelmColumn` and `LabelColum`
++ core: Pass the `Root` as owned parameter (without reference) to `Component::init`
 + core: Add `visible_on_activate()` to `RelmApp` to prevent the app window from being visible immediately
 + core: Make `into_stream()` method on `Receiver` public
 + examples: Add libadwaita `Toast` example
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ impl SimpleComponent for App {
     // Initialize the component.
     fn init(
         counter: Self::Init,
-        root: &Self::Root,
+        root: Self::Root,
         sender: ComponentSender<Self>,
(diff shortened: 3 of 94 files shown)


hwittenborn (changes requested):
Should these spots also take an owned root?

https://github.com/Relm4/Relm4/blob/owned-root-init/relm4/src/component/async/traits.rs#L52
https://github.com/Relm4/Relm4/blob/owned-root-init/relm4/src/factory/async/traits.rs#L55
https://github.com/Relm4/Relm4/blob/owned-root-init/relm4/src/factory/async/traits.rs#L63

AaronErhardt:
> Should these spots also take an owned root?

Good catch. I've added a new commit to change the missing occurrences.

AaronErhardt:
@hwittenborn could you review the latest changes? Since this touches a lot of code it would be nice to have this merged to avoid merge conflicts.

hwittenborn:
Sorry about the way @AaronErhardt, I got distracted with school and some other projects and complete forgot to get this looked over.

I'm pretty good on free time now though, I'm just busy traveling until around 7 pm tonight. I can get this reviewed right after that though.

AaronErhardt:
Thanks for the quick reply! No problem. It's not urgent, but it would be nice to have this merged rather soon of course.

hwittenborn:
Yeah for sure, I don't particularly like waiting on PRs for a long time in my own projects so I definitely know the desire to keep things moving forward. I'll get this look at soon 👍

hwittenborn (changes requested):
Just found one more thing before this looks good to me:

It looks like this line still needs to be changed: https://github.com/Relm4/Relm4/blob/owned-root-init/relm4-components/examples/web_image.rs#L68. I'm not sure why the CI didn't catch that, but it looks like adding `--all-features` (or `--features web`) to https://github.com/Relm4/Relm4/blob/main/.github/workflows/rust.yml#L55-L56 would fix it up.

---
This doesn't have to be addressed in this PR, but I'm wondering if we should consider moving the `root` parameter in things like [`AsyncComponent::update`](https://docs.rs/relm4/latest/relm4/component/trait.AsyncComponent.html#method.update) to containing an owned `Self::Root` as well. I think that'd be best for keeping the API consistent, but I'd be open to getting that discussed after this PR if need be too.

hwittenborn:
Hey @AaronErhardt, any updates on this? I know you said you'd like to get this merged since so much relies on it, just don't want things in the repository to end up in a clunky state or anything.

AaronErhardt:
@hwittenborn I had an unfinished path laying around in my local git repo. Simply forgot to finish it due to being busy with other things. The code should be correct now and the CI should catch future mistakes that only appear with certain features enabled.

PR #568, opened 2023-10-26
Add adw::Toast example

sgued:
#### Summary

This PR adds an example for using an `adw::Toast`.

The current example looks pretty bad since I'm just figuring it out (which is why I wished there was already such an example and want to add my own). This PR is a much a PR to improve the examples as it is me asking for help to figure it out.

#### To solve

- [x] I couldn't find a way to make the button work without having to create the `ActionGroup`. Is that really necessary? It feels really weird as an API and I don't understand what it brings (I am not at all familiar with `gio` and `Gtk` overall).
- [x] I don't understand how to access the `toast_overlay` from the `update` method. By naming it I can get it from the widgets, but I don't see how to get the widgets from `update`, so I instead put it in the model, but that again feels out of place (plus it's weird since the model needs to be created before the widgets).

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/examples/libadwaita/Cargo.toml
+++ b/examples/libadwaita/Cargo.toml
@@ -20,3 +20,8 @@ path = "tab_factory.rs"
 [[example]]
 name = "tab_game"
 path = "tab_game.rs"
+
+[[example]]
+name = "toast"
+path = "toast.rs"
+required-features = ["relm4/gnome_43"]
--- a/examples/libadwaita/toast.rs
+++ b/examples/libadwaita/toast.rs
@@ -1,0 +1,109 @@
+use gtk::prelude::*;
+use relm4::prelude::*;
+use std::cell::Cell;
+
+struct App {
+    activated: &'static str,
+    toast: Cell<Option<adw::Toast>>,
+}
+
+#[derive(Debug)]
+enum Msg {
+    Activate,
+    Cancel,
+}
+
+#[relm4::component]
+impl SimpleComponent for App {
+    type Init = ();
+    type Input = Msg;
+    type Output = ();
+
+    view! {
+        adw::Window {
+            set_title: Some("Simple app"),
+            set_default_size: (300, 200),
+
+            gtk::Box {
+                set_orientation: gtk::Orientation::Vertical,
+
+                adw::HeaderBar {
+                    #[wrap(Some)]
+                    set_title_widget = &adw::WindowTitle {
+                        set_title: "Toast",
+                    }
+                },
+
+                #[name = "toast_overlay"]
+                adw::ToastOverlay {
+                    set_vexpand: true,
+                    gtk::Box {
+                        set_orientation: gtk::Orientation::Vertical,
+                        set_spacing: 5,
+                        set_margin_all: 5,
+                        set_valign: gtk::Align::Center,
+
+                        gtk::Button {
+                            set_hexpand: false,
+                            set_vexpand: false,
+                            set_label: "Activate",
+                            connect_clicked => Msg::Activate,
+                        },
+                        gtk::Button {
+                            set_hexpand: false,
+                            set_vexpand: false,
+                            set_label: "Cancel",
+                            connect_clicked => Msg::Cancel,
+                        },
+                        gtk::Text {
+                            #[watch]
+                            set_text: model.activated,
+                        },
+                    },
+
+                    #[watch]
+                    add_toast?: model.toast.take(),
+                }
+            }
+
+        }
+    }
+
+    // Initialize the component.
+    fn init(
+        _: Self::Init,
+        root: &Self::Root,
+        sender: ComponentSender<Self>,
+    ) -> ComponentParts<Self> {
+        let model = App {
+            activated: "Idle",
+            toast: Cell::new(None),
+        };
+        // Insert the code generation of the view! macro here
+        let widgets = view_output!();
+
+        ComponentParts { model, widgets }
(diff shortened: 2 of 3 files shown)


AaronErhardt:
>  I couldn't find a way to make the button work without having to create the ActionGroup. Is that really necessary? It feels really weird as an API and I don't understand what it brings (I am not at all familiar with gio and Gtk overall).

You are probably looking for [`connect_button_clicked()`](https://world.pages.gitlab.gnome.org/Rust/libadwaita-rs/stable/latest/docs/libadwaita/struct.Toast.html#method.connect_button_clicked)

> I don't understand how to access the toast_overlay from the update method. By naming it I can get it from the widgets, but I don't see how to get the widgets from update, so I instead put it in the model, but that again feels out of place (plus it's weird since the model needs to be created before the widgets).

This is indeed a non-trivial problem. Relm4, being inspired by Elm, separates UI and app state, but this doesn't work well for toasts in particular.

The best solution I can think of would be to store `Cell<Option<Toast>>` in the model and then place a `Some(Toast::new(...))` in there when needed. In the view macro, you'd then use `add_toast?: model.toast.take(),` and `#[update]` to set the toast whenever it was set during the `update` method.

Does that make sense for you?

examples/libadwaita/Cargo.toml
```diff
@@ -8,6 +8,7 @@ edition = "2021"
 rand = "0.8.5"
 relm4 = { path = "../../relm4", features = ["libadwaita"] }
 tokio = { version = "1.25", features = ["time"] }
+adw = { version = "0.5", package = "libadwaita" }
```
AaronErhardt, line 11: This is not needed btw. because Relm4 with the "libadwaita" feature re-exports it, so you can just use `use relm4::adw;` to have `adw` in scope.

sgued: It's not added for the `asw` import, but for the `adw/v1_2` required feature in the example.

examples/libadwaita/toast.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    fn update(&mut self, msg: Self::Input, sender: ComponentSender<Self>) {
+        match msg {
+            Msg::Activate => {
+                self.activated = "Active";
+                let toast = adw::Toast::new("Activated");
+                toast.set_action_name(Some("toast.cancel"));
+                toast.set_button_label(Some("Cancel"));
+                toast.set_timeout(0);
+                toast.connect_button_clicked(move |this| {
```
sgued, line 103: Here I do already use the `connect_button_clicked`, but unless I do all the activation stuff for the action the button show up grey-ish like if it's deactivated and can't be clicked.

examples/libadwaita/Cargo.toml
AaronErhardt: Oh, I didn't see that. Yet, you could also use "relm4/gnome_43" instead, which will also enable the required libadwaita feature.

sgued:
> The best solution I can think of would be to store `Cell<Option<Toast>>` in the model and then place a `Some(Toast::new(...))` in there when needed. In the view macro, you'd then use `add_toast?: model.toast.take(),` and `#[update]` to set the toast whenever it was set during the `update` method.
> 
> Does that make sense for you?

That's exactly the kind of thing I was looking for thanks!

examples/libadwaita/toast.rs
AaronErhardt: Weird. I'll look into this myself. I don't think that's supposed to happen because otherwise the method would be kind of pointless.

examples/libadwaita/Cargo.toml
sgued: I will do that

examples/libadwaita/toast.rs
AaronErhardt: I don't see a difference, to be honest. It looks correct for me regardless of whether actions where used or not.

Without actions:

Screenshot from 2023-10-26 23-15-11

With actions:

Screenshot from 2023-10-26 23-15-46

I think it's just the theme that makes it look disabled because of the low contrast. The button in the Adwaita Demo app also looks like this:

image

sgued: Light theme with action stuff:

image

Light theme without action stuff, **the button can't be clicked**:

image


Dark theme with action stuff:

image

Dark theme without action stuff:

image


I'm on Arch with  libadwaita 1:1.4.0-1, gtk4 1:4.12.3-2

AaronErhardt: Weird, I also have libadwaita-1.4.0-1 (but gtk4 4.12.3-**1**) on Fedora 39 and even if I remove the entire action stuff and the signal handler (leaving just the label and the timeout) the button is sensitive and I can click it. I don't think you're doing something wrong, this seems to be an libadwaita issue...

sgued: I will report it upstream then.

sgued: Ok, I actually got it wrong. I forgot to remove the `set_action_name`. Now it works properly.

sgued:
I pushed fixes following your feedback, now it looks to me like it's ready for inclusion.

The pipeline appears to fail for a reason that is not relevant to this PR.

AaronErhardt (approved):
Looks good. Thanks!

PR #569, opened 2023-10-30
Move internal initialization to startup signal handler

azymohliad:
#### Summary

Currently `RelmApp` does full components initialization in `gtk::Application::activate` signal handler. This is problematic for apps that needs to be able to run in the background, because they use `activate` signal to switch back to foreground. This PR moves the initialization to `startup` signal handler, and uses `activate` signal only for setting the main window visible.

This shouldn't change the behavior of any existing apps unless they used `gtk::Application::connect_startup` directly.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
++ core: Move internal inititalization to `gtk::Application::startup` signal handler
 + examples: Add libadwaita `Toast` example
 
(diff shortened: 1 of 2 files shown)


AaronErhardt (approved):
Looks good. Thanks!

azymohliad:
Thank you!

PR #577, opened 2023-11-21
Increase flexibility of `Alert` component

### LINKED ISSUE #575: Increase flexibility of `Alert` component
I'm wanting to use the `Alert` component from `relm4_components`, but it currently isn't flexible enough for my needs.

If I were to use the `Alert` component, I'd need the following functionality in it:

- The ability to only have one button. I have a use case where I'm wanting to show an error to the user, and only have them click an `Ok` button.
- The ability to manually hide the widget. I have another use case where the user will have a `Cancel` button on the `Alert` component to quit a task, but I also want to close it manually if that task finishes before they do such.
- The ability to change the text on any of the widget's settings after creation. I'm wanting to store this widget as part of it's parent widget, but I won't know the text I'm displaying to the user until a later point in my app's execution cycle (I'll be displaying a `Logging in to {cloud provider}` message as part of [hwittenborn/celeste](https://github.com/hwittenborn/celeste)).

I'm willing to submit a PR for these changes, I'm just seeing if it's something y'all would be open to accepting.

hwittenborn:
#### Summary
The `Alert` component from `relm4_components` previously didn't have the flexibility I needed for certain uses cases (as discussed in #575). This PR adds those changes.

I also added a `Default` implementation to aid in quickly creating new instances of `Alert` when a ton of configuration isn't desired (i.e. a title and single button, and then `..Default::default()` can be called instead of specifying everything else manually). I added documentation on how the `Default` implementation behaves as I currently don't have it using the `Default` values for all the struct's fields.

As part of my testing I also noticed that hitting the `Esc` key while an `Alert` dialog is open will make it unable to be shown again. I'm not sure what the best approach to fixing that is, but I thought I'd mention it (I can make an issue if it should be tackled outside of this PR).

I also wasn't able to add the extra custom widget functionality mentioned in #575, as there wasn't a clean way to do it while still preserving the ability to modify the struct's settings. `adw::MessageDialog` has some utilities to do such, but that would make the widget require libadwaita and I wasn't sure if y'all wanted that (I would definitely prefer it, any thoughts?).

Closes #575.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@
 
 + core: Move internal initialization to `gtk::Application::startup` signal handler
 + core: Remove deprecated RelmApp methods
++ components: Increase flexibility of `Alert` component
 
 ### Fixed
 
--- a/relm4-components/examples/alert.rs
+++ b/relm4-components/examples/alert.rs
@@ -108,8 +108,8 @@ impl SimpleComponent for App {
                 .launch(AlertSettings {
                     text: String::from("Do you want to quit without saving? (First alert)"),
                     secondary_text: Some(String::from("Your counter hasn't reached 42 yet")),
-                    confirm_label: String::from("Close without saving"),
-                    cancel_label: String::from("Cancel"),
+                    confirm_label: Some(String::from("Close without saving")),
+                    cancel_label: Some(String::from("Cancel")),
                     option_label: Some(String::from("Save")),
                     is_modal: true,
                     destructive_accept: true,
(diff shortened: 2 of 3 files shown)


hwittenborn:
Is that GTK error related to something I changed?

hwittenborn:
Nevermind it looks like it's happening on master too, is the image from https://github.com/Relm4/docs just out of date?

It looks like it was using automatic builds previously, is there a reason they got stopped?

AaronErhardt (approved):
Looks good. I don't know what happened to the CI. The relm4-docs image is up-to-date, maybe the gtk-rs image on which it is based on has a problem...

PR #579, opened 2023-11-22
core: Add WIP panic handler code

### LINKED ISSUE #578: Add option to not capture panics
It appears that when a panic happens anywhere in the `Component` trait (or any others that function similarly to it) that panics are captured. After the panic occurs, it seems like certain parts of my application (that I haven't seen a pattern to yet) are no longer running, and then my application is left in an unclean state where it can't really do much without breaking.

I'm not sure if this is intended behavior, but I'd really like it if I could at least have an option to disable this panic capturing.

AaronErhardt:
#### Summary

Fixes #578
@hwittenborn

#### Checklist

- [ ] cargo fmt
- [ ] cargo clippy
- [ ] cargo test
- [ ] updated CHANGES.md


hwittenborn:
We sort of discussed this in that issue and in the Matrix room, but what's the benefit of this over just capturing panics and aborting on a crate-wide level?

AaronErhardt:
With the approach in this PR you can specify individual behavior for each component. You can choose what should happen and whether the entire app should be aborted or not. Also, it's just a method on the `Component` trait instead of the panic hook, which most Rust users don't even know about.

hwittenborn:
> You can choose what should happen and whether the entire app should be aborted or not.

That was leading me to think about when you would ever *not* want the entire app to be aborted (or at least start unwinding, which isn't a current possibility). When a component panics you can't really use it much anymore, and there doesn't seem to be a good way to handle that kind of state from parent components.

I feel like the lack of handling in that regard from parent components would just make things bound to break, as components don't expect that their child components are just going to panic. I know at least in the app that I'm working on that I'm not expecting things to work at all once a panic occurs, and the pre-Relm4 code where panics weren't captured behaved in a similar manner.

AaronErhardt:
Yes, in most cases you want to stop the entire app, so that's the default behavior introduced by this PR. You can still return `false` from the `handle_panic()` method if you think you can recover from the panic.

AaronErhardt:
Let's say you use some crate which has a bad API that panics instead of returning an error (plotters would be a prominent example). You could a worker or component that manages the export of the plot to a file and if a panics happens you can catch it and continue the execution.

Or you want to execute some additional code, for example for recording the application state and only then you want to actually shut down the app.

hwittenborn:
I'm still a bit questionable on how needed this should be, I feel like there isn't really a case where it provides much use.

> You could a worker or component that manages the export of the plot to a file and if a panics happens you can catch it and continue the execution.

I feel like this kind of stuff should be handled in application code itself, e.g. via the use of [`catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) followed by sending an output message to the component's parent about the error. Something like this dummy code is what I'm imagining would be the best way to tackle these kind of issues:

```rust
enum AppOutput {
    ...
    PanicErr(Err)
}

struct AppModel;

impl Component for AppModel {
    ...

    fn update(...) {
        if let Err(err) = catch_unwind(|| { ... }) {
        sender.output(AppOutput::PanicErr(err));
        }
    }
```

And then any panics outside of that case should just crash the app in my opinion. I can't really see a useful case outside of that.

AaronErhardt:
I think you're right, catching a panic should be handled by the user manually. I've updated the PR now accordingly and extended the functionality to factories.

hwittenborn (changes requested):
Sorry for the late review, I've been quite busy recently :/, things should start to get back to normal now though. I left some comments on things I noticed; I haven't looked over everything in this PR yet, but it's mostly functioning pretty much exactly like what I originally wanted.

hwittenborn:
The only other thing I'm curious about is a way panics could be propagated all the way back to `RelmApp::run` or `RelmApp::run_async`. I'm not sure how that could be done with the usage of async code, but the current implementation is still better compared to what happens currently.

PR #586, opened 2023-12-10
core: Improve late init callbacks after startup

### LINKED ISSUE #584: `transient_for` errors in child component
I have the following hierarchy: `App > Component` and `Component::init` crates `OpenDialog`.

The issue is that in this case `transient_for(&root)` results in this:
```
2023-12-09T12:40:47.898031Z ERROR relm4::component::sync::builder: Couldn't find root of transient widget
```

The reason is that `Component` wasn't yet added to window, we're still building it. I didn't find clearer solution than to pass `Window` down as `Component::Init`, but it feels like that error shouldn't happen and `OpenDialog` should be able to attach to window once window is available even if is not available during construction.

AaronErhardt:
#### Summary

Should fix #584

@nazar-pc can you try this branch? I'm not entirely sure whether this fixes your specific issue, but this change should be good in general.

With this PR, the following order of operations should start working:

1. Build a component
2. Call `transient_for` with a widget that is not connected yet
4. Add the widget to a window (or any other widget somehow connected to a window)
5. Return (from `init`, `update` or similar)

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,7 @@
 
 + core: Report the correct dimensions in DrawingHandler when a scaling factor is set
 + core: Setting the visibility of the main window isn't overridden by `RelmApp` anymore
++ core: Fix timing of `transient_for()` when called after app init
 
 ## 0.7.0-beta.2 - 2023-10-14
 
--- a/relm4/src/late_initialization.rs
+++ b/relm4/src/late_initialization.rs
@@ -11,10 +11,13 @@ pub(super) fn register_callback(func: Callback) {
     if let Some(inner) = Lazy::get(&LATE_INIT) {
         // If `Lazy` was initialized and is not empty,
         // this means that `run_late_init` has run already.
-        // In this case, we need to call the callback immediately,
-        // otherwise it will never be executed.
+        // In this case, we call the callback as soon as the current function
+        // returns and thereby yields to the glib runtime.
         if inner.get().borrow().is_empty() {
-            return func();
+            crate::spawn_local_with_priority(gtk::glib::Priority::HIGH, async {
+                func();
+            });
+            return;
         }
     }
     LATE_INIT.get().borrow_mut().push(func);


nazar-pc (approved):
No error in logs with this change, so I think it works

nazar-pc:
I should add that I was testing this with a synchronous component, but I don't think it will work with an async component, am I correct?

nazar-pc:
I think solution might be in case top most node is not `Window` to subscribe to an event when it that widget is attached to something and traverse the tree further until `Window` is finally reached.

Another edge-case that I think is not handled right now is when dialog is set as transient for some widget, but then that widget is moved to a different window (it is possible to do so, right?), then dialog will remain a transient of a wrong window after such thing happens.

AaronErhardt:
> I should add that I was testing this with a synchronous component, but I don't think it will work with an async component, am I correct?

I think it should. It should work in general as long as you attach the root widget of a child component within your `init()` function. Only if you `await` between calling `transient_for` and attaching the widget this would still not work.

> I think solution might be in case top most node is not Window to subscribe to an event when it that widget is attached to something and traverse the tree further until Window is finally reached.

I thought there is no such event, but [`connect_parent_notify`](https://gtk-rs.org/gtk4-rs/git/docs/gtk4/prelude/trait.WidgetExt.html#method.connect_parent_notify) might work. I'll investigate this further.

> Another edge-case that I think is not handled right now is when dialog is set as transient for some widget, but then that widget is moved to a different window (it is possible to do so, right?), then dialog will remain a transient of a wrong window after such thing happens.

Yes, but only if you do this after `init`, at which point this is not reasonable to take this into consideration from the side of Relm4.

AaronErhardt:
>> I think solution might be in case top most node is not Window to subscribe to an event when it that widget is attached to something and traverse the tree further until Window is finally reached.

> I thought there is no such event, but [connect_parent_notify](https://gtk-rs.org/gtk4-rs/git/docs/gtk4/prelude/trait.WidgetExt.html#method.connect_parent_notify) might work. I'll investigate this further.

It turns out that doesn't work reliable. The widget could be attached to another widget, but that widget has no window yet. We could move up to the "highest" widget without attachment and wait for it to be attached, but that attachment does not guarantee that now there is a window available at the top. Overall, I think the current approach is the best solution.

Also, the code should be ready now.

nazar-pc:
Can you rebase this on `main`? It seems to be quite a bit behind.

nazar-pc (commented):
This works for synchronous components, I just verified that. But I think it will not work for async components that actually wait for something and will definitely not work for components that are constructed and stored in a variable, while only being attached to a window at a much later point in the future.

nazar-pc:
FYI clippy failed in CI and still fails on `main`, I think that recursion needs to be fixed

AaronErhardt:
I think that's a false positive. The code looks correct to me, there's no recursion. Unfortunately, it's common for new clippy lint's to create false positives. We've had that several times before and even disabled nightly or beta CI once just because of that.

nazar-pc:
Looks like it was fixed already (https://github.com/rust-lang/rust-clippy/issues/12133), is some caching involved in CI?

AaronErhardt:
Maybe the fix took some time to propagate to nightly. The CI is fixed now without changing this specific issue.

PR #589, opened 2023-12-12
Column resize/expand functionality

### LINKED ISSUE #588: Resizeable and expandable columns with TypedColumnView
GTK has the ability to make columns resizeable, as seen [here](https://docs.rs/gtk4/0.7.3/gtk4/struct.ColumnViewColumn.html#method.set_resizable) in the Rust bindings for GTK4, and [here](https://docs.gtk.org/gtk4/method.ColumnViewColumn.set_resizable.html) in the GTK4 documentation. However, Relm4's `TypedColumnView` does not seem to expose that functionality. This feature would be very useful for the program I'm currently creating.

Similarly, there is also a setting that causes the columns to expand to take up all available space ([Rust docs](https://docs.rs/gtk4/0.7.3/gtk4/struct.ColumnViewColumn.html#method.set_expand), [GTK docs](https://docs.gtk.org/gtk4/method.ColumnViewColumn.set_expand.html)). This would also be quite useful for my app.

Could these features be exposed in the Relm4 bindings?

Lynnesbian:
#### Summary

Adds `const bool`s to [`RelmColumn`](https://docs.rs/relm4/0.7.0-beta.2/relm4/typed_view/column/trait.RelmColumn.html) (and [`LabelColumn`](https://docs.rs/relm4/0.7.0-beta.2/relm4/typed_list_view/trait.LabelColumn.html)) that allow for setting the created columns to be resizeable, or to automatically expand.

Fixes #588 

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/relm4/src/typed_view/column.rs
+++ b/relm4/src/typed_view/column.rs
@@ -26,6 +26,10 @@ pub trait RelmColumn: Any {
 
     /// The columns created for this list item.
     const COLUMN_NAME: &'static str;
+    /// Whether to enable resizing for this column
+    const ENABLE_RESIZE: bool = false;
+    /// Whether to enable automatic expanding for this column
+    const ENABLE_EXPAND: bool = false;
 
     /// Construct the widgets.
     fn setup(list_item: &gtk::ListItem) -> (Self::Root, Self::Widgets);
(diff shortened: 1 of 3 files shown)


relm4/src/typed_view/column.rs
```diff
@@ -238,6 +248,9 @@ where
         let sort_fn = C::sort_fn();
 
         let c = gtk::ColumnViewColumn::new(Some(C::COLUMN_NAME), Some(factory));
+        c.set_resizable(C::ENABLE_RESIZE);
+        c.set_expand(C::ENABLE_EXPAND);
+        c.set_resizable(C::ENABLE_RESIZE);
```
AaronErhardt, line 253: Is there a reason why `set_resizable` is called twice?

Lynnesbian: That's a mistake on my part, let me fix it now. Thanks for pointing it out.

AaronErhardt (approved):
Thanks! Everything looks good now.

PR #593, opened 2023-12-25
WIP: multiple application windows

vhdirk:
#### Summary

This PR is a draft to allow gtk::Application to be used as root element, paving the way for multiple application windows.

Currently, it is more a POC than anything else. It shuts down instantly and always shows the message:

``` 
(application:296406): Gtk-CRITICAL **: 23:13:26.648: New application windows must be added after the GApplication::startup signal has been emitted.
```

#### Checklist

- [ ] cargo fmt
- [ ] cargo clippy
- [ ] cargo test
- [ ] updated CHANGES.md

### DIFF
--- a/examples/application_async.rs
+++ b/examples/application_async.rs
@@ -1,6 +1,7 @@
+use async_trait::async_trait;
 use gtk::prelude::*;
 use relm4::*;
-use relm4::factory::FactoryVecDeque;
+use relm4::factory::{AsyncFactoryVecDeque, AsyncFactoryComponent};
 use relm4::prelude::*;
 
 struct Window {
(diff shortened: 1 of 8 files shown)


relm4/src/app.rs
```diff
@@ -169,6 +170,55 @@ impl<M: Debug + 'static> RelmApp<M> {
...
+            app,
+            broker,
+            args,
+            ..
+        } = self;
+
+        let hold_guard = Rc::new(RefCell::new(Some(app.hold())));
+        app.connect_window_added(move |_app, _window| {
+            hold_guard.borrow_mut().take();
+        });
```
AaronErhardt, line 189: I think `Rc` is not necessary here and `Cell` should be used instead of `RefCell`. Same for the async version later.

examples/application_async.rs
```diff
@@ -0,0 +1,126 @@
...
+        match msg {
+            AppMsg::Activate(init) => {
+                windows_guard.push_back(init);
+            }
+        }
+    }
+}
+
+fn main() {
+    let app = RelmApp::new("relm4.example.application");
```
AaronErhardt, line 124: This isn't really important, but for consistency, the app id should end with async here.

AaronErhardt (changes requested):
Looks good overall. Does it still not work for async?

PR #614, opened 2024-02-25
Implement TypedGridView

tsmweb:
Typed wrapper implementation for gtk::GridView. Largely based on TypedListView.


AaronErhardt (changes requested):
Looks good, but it would be nice if rustfmt and clippy were fixed and an entry to CHANGES.md was added.

AaronErhardt (approved):
Looks good, thanks!

PR #628, opened 2024-04-19
Add `#[derive(Clone)]` for actions created by macros

lessneek:
#### Summary
Add `#[derive(Clone)]` for actions created by macros in order to have a cloning support of `RelmAction<>`.

The problem discussion: https://github.com/orgs/Relm4/discussions/599

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Changed
+
++ core: Add `#[derive(Clone)]` for actions created by macros
+
 ## 0.8.1 - 2024-3-13
 
 ### Fixed


AaronErhardt:
I don't think this is the right approach. The problem is rather caused by the `#[derive(Clone)]` on `RelmAction`, which adds a bound to the action name to also implement `Clone`. Because the action name is just a generic parameter, that is not necessary and if we add a proper `Clone` implementation manually, that would fix the problem.

Implementing `Clone` on the macro generated types also fixes the issue, but we don't need those trait implementations actually, so I'd prefer the other solution, see #629. What do you think?

PR #632, opened 2024-04-30
core: Add `allow_multiple_instances()` to `RelmApp`

### LINKED ISSUE #396: Relm4 should support multiple application instances
We currently panic if multiple instances of the same application are started:

https://github.com/Relm4/Relm4/blob/9ad0a5dabae5db0ca88ad7fa791d533c7bd32a2a/relm4/src/app.rs#L87

This is very limiting for certain classes of application (file managers, for instance). We should remove this limitation and handle this case gracefully.

falko17:
#### Summary
Usually, starting an already-running application again causes nothing to happen. However, with `allow_multiple_instances(true)`, the `NON_UNIQUE` flag is set, which makes it possible to run multiple instances of the application to run at the same time.
This should fix #396.

#### Checklist

- [X] cargo fmt
- [X] cargo clippy
- [X] cargo test
- [X] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
++ core: Add `allow_multiple_instances()` to `RelmApp` to allow multiple concurrent instances of the same application
+
 ### Fixed
 
 + core: Don't require `Clone` and `Debug` for the generic action name parameter in `RelmAction`
--- a/relm4/src/app.rs
+++ b/relm4/src/app.rs
@@ -84,6 +84,22 @@ impl<M: Debug + 'static> RelmApp<M> {
         self
     }
 
+    /// If `true`, allow multiple concurrent instances of the application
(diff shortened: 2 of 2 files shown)


relm4/Cargo.toml
```diff
@@ -42,6 +42,7 @@ adw = { version = "0.6", optional = true, package = "libadwaita" }
 flume = "0.11.0"
 futures = "0.3.30"
 fragile = "2.0.0"
+gio = "0.19"
```
AaronErhardt, line 45: Including `gio` is not necessary because it is re-exported under `gtk::gio`: https://gtk-rs.org/gtk4-rs/git/docs/gtk4/index.html#reexports

falko17: Oh, I missed that, thanks. Fixed now.

AaronErhardt (approved):
Looks good, thanks!

PR #642, opened 2024-05-27
examples: drop sub-components instead of hiding

blabno:
#### Summary

Add Example for Sub-Components Drop Instead of Hide

This pull request introduces an example demonstrating how to make sub-components drop instead of merely hiding them.

Benefits:
* Dropping sub-components allows for better resource cleanup.
* When a component is removed from its parent, it no longer impacts the parent's size, unlike hiding which can leave hidden elements occupying space.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/examples/drop_sub_components.rs
+++ b/examples/drop_sub_components.rs
@@ -1,0 +1,267 @@
+use std::fmt::Debug;
+
+use relm4::gtk::prelude::*;
+use relm4::prelude::*;
+use relm4::RelmRemoveAllExt;
+
+mod css {
+    pub const COMPONENT_SPACING: i32 = 5;
+}
+
+enum AppMode {
+    Initial(#[allow(dead_code)] Controller<InitialScreen>),
+    SubScreen1(#[allow(dead_code)] Controller<SubScreen1>),
+    SubScreen2(#[allow(dead_code)] Controller<SubScreen2>),
+}
+
+struct App {
+    mode: Option<AppMode>,
+}
+
+#[derive(Debug)]
+enum Msg {
+    ShowInitialScreen,
+    ShowSubScreen1,
+    ShowSubScreen2,
+}
+
+#[relm4::component]
+impl Component for App {
+    type Init = ();
+    type Input = Msg;
+    type Output = ();
+    type CommandOutput = ();
+
+    view! {
+        gtk::Window {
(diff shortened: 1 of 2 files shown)


examples/drop_sub_components.rs
```diff
@@ -0,0 +1,280 @@
...
+        match message {
+            Msg::ShowInitialScreen => {
+                let controller =
+                    InitialScreen::builder()
+                        .launch(())
+                        .forward(sender.input_sender(), |msg| match msg {
+                            InitialScreenOutput::ShowSubScreen1 => Msg::ShowSubScreen1,
+                            InitialScreenOutput::ShowSubScreen2 => Msg::ShowSubScreen2,
+                        });
+                widgets.container.append(controller.widget());
```
AaronErhardt, line 86: Does that work as expected? I think the component runtime will be dropped as expected, but the widgets will remain in the `Box`. Using [`CenterBox::set_center_widget`](https://gtk-rs.org/gtk4-rs/git/docs/gtk4/struct.CenterBox.html#method.set_center_widget) would automatically remove the widgets as well.

examples/drop_sub_components.rs
```diff
@@ -0,0 +1,280 @@
...
+            set_hexpand: true,
+            set_vexpand: true,
+            set_halign: gtk::Align::Center,
+            set_valign: gtk::Align::Center,
+            gtk::Box {
+                set_orientation: gtk::Orientation::Vertical,
+                set_spacing: css::COMPONENT_SPACING,
+                gtk::Button {
+                    set_label: "Sub screen 1",
+                    connect_clicked[sender] => move |_| sender.output(InitialScreenOutput::ShowSubScreen1).unwrap_or_else(|e| log_sender_output_error(INITIAL_SCREEN_COMPONENT_NAME, e))
```
AaronErhardt, line 147: Using `.unwrap()` or `.expect()` might be ok in this case. If the parent component forgets to forward the output messages, that's a considerable programming error that would justify a panic IMO.

AaronErhardt (changes requested):
Thanks! Code looks good overall, I just noticed some minor things.

AaronErhardt (approved):
Looks good, thanks!

PR #648, opened 2024-06-03
Add global functions to set CSS

hwittenborn:
#### Summary
Part of the work for #645. As discussed in the Matrix dev room, feel free to not merge this until we look at how it's used in the PR for that (once I have it created).

#### Checklist

- [x] cargo fmt
- [ ] cargo clippy
- [ ] cargo test
- [x] updated CHANGES.md

PR #649, opened 2024-06-03
Rewrite `Alert` component from stratch to work with both GTK4 and Adwaita

### LINKED ISSUE #645: Rewrite `Alert` component to avoid using `MessageDialog`
Currently, the `Alert` component uses `gtk::MessageDialog`, which causes an issue when you want the body text of the dialog to be center-aligned. The `adw::MessageDialog` centers such text, which - while minor - is desirable by people who want their apps to follow the Adwaita styling.

There's the option of just using `adw::MessageDialog`, which I've currently [mostly implemented](https://github.com/hwittenborn/Relm4/commit/887d03fd06eb7eab34e7ed5fcba1386533e14c40). I'm wondering if it would be better to make a widget that functions the same as `adw::MessageDialog` though, but is just written from scratch with only GTK4. This would allow more flexibility to be put on the widget (such as how text is aligned), and (hopefully) allow backward compatibility to older GTK4 versions.

Any thoughts? I'm open to creating a PR for the Adwaita alert component I've made, but I feel like it's a bit hacky since the Adwaita-based component is largely the same as the GTK4 one. I've also managed to add in support for an extra widget on the Adwaita component, and having a feature mismatch between the two didn't make a ton of sense to me.

hwittenborn:
*This PR needs to be merged after #648, as it builds on top of changes from that one.*

#### Summary
Resolves #645. I did some tests to make sure everything looks correct:

**GTK4:**
image

**Adwaita:**
image

When writing the [original Adwaita-based version](https://github.com/hwittenborn/Relm4/commit/887d03fd06eb7eab34e7ed5fcba1386533e14c40) that would've resulted in two separate `Alert` components, I had added some extra features (such as adding in an extra widget). I haven't added such functionality here yet to keep the changes minimal, but if it feels appropriate to do so I'm fine adding that in (or in a separate PR, either is fine by me).

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


hwittenborn:
Also of note - GTK4's `MessageDialog` doesn't have the text center-aligned:
image

However, I made it to where [the text is center aligned](https://github.com/hwittenborn/Relm4/blob/235bc95cd61b9cf5b0b94be4df27ce6c5fbc572a/relm4-components/src/alert/mod.rs#L130) (which is how Adwaita does it):
image

I'm not sure if this is desired, but it was at least by me where the left-aligned text felt off with everything else being center-aligned.

relm4-components/src/alert/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+use gtk::prelude::{BoxExt, ButtonExt, GtkWindowExt, OrientableExt, WidgetExt};
+use relm4::{gtk, Component, ComponentParts, ComponentSender, RelmWidgetExt};
+
+static LIBADWAITA_ENABLED: bool = cfg!(feature = "libadwaita");
+static COMPONENT_CSS: &str = include_str!("style.css");
+
+static ERROR_CSS: &str = "error";
+static TITLE_CSS: &str = "title-2";
+static DESTRUCTIVE_CSS: &str = "destructive-action";
+static FLAT_CSS: &str = "flat";
```
AaronErhardt, line 14: Shouldn't is be `const` instead of `static`?

relm4-components/src/alert/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    fn init(
+        settings: AlertSettings,
+        root: Self::Root,
+        sender: ComponentSender<Self>,
+    ) -> ComponentParts<Self> {
+        relm4::set_global_css_with_priority(
+            COMPONENT_CSS,
+            gtk::STYLE_PROVIDER_PRIORITY_APPLICATION,
+        );
```
AaronErhardt, line 214: I'm not sure how GTK handles this, but ideally we would call this only the first time an `AlertComponent` is created.

relm4/src/app.rs
AaronErhardt, line 1: I guess a rebase against main would be nice to avoid those changes in the diff.

AaronErhardt (changes requested):
Looks very nice! I just found a few minor things to improve.

relm4-components/src/alert/mod.rs
hwittenborn: I chose `static`, as otherwise the string gets inlined everywhere it's used. Would you rather it not behave like that?

After #650 gets merged, these shouldn't be required anyway though. If you'd prefer things to be `const` I'll need to get things changed up before that PR gets merged as well.

relm4-components/src/alert/mod.rs
hwittenborn: I'd assume it would handle it the same way as if we override a styling that already exists (like how a custom theme might do). The styling shouldn't need to be initialized more than once in this case anyway though.

relm4/src/app.rs
hwittenborn: Will do, I'll get that done in a second.

relm4-components/src/alert/mod.rs
AaronErhardt: I prefer `const` when possible. I don't think rustc will inline `const` items everywhere, rather `const` items are not guaranteed to have a certain memory location. Since we don't need `static` in the sense of a global variable here and rather define constants, `const` should be idiomatic (and maybe even slightly more performant).

hwittenborn: [It does look like Rust inlines `const` usage](https://doc.rust-lang.org/std/keyword.const.html):

> ...constants are inlined wherever they’re used, making using them identical to simply replacing the name of the `const` with its value.

I'm not sure if that's a problem for you, but I thought I'd mention it just in case. If that's fine, I'm good to move them over to using `const`.

AaronErhardt: They might be inlined, but the compiler will not create unnecessary space for them. In fact, I just checked out of curiosity and the compiler will emit the exact same assembly regardless of `static` or `const` being used, with the only difference being the name of the memory region it creates for the constant value. So since `const` is semantically more correct, I'd simply use that.

hwittenborn: I wasn't aware of that, I guess the docs misled me a bit :p. Anyhow, I can't argue with what's proven, so I'll go ahead and get everything moved over.

AaronErhardt (approved):
Thanks!

PR #657, opened 2024-07-12
components: Add `icon` to `OpenButton`

tymmesyde:
#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
++ components: Add `icon` to `IconButton`
 
 ## 0.9.0 - 2024-7-12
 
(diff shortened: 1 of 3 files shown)


AaronErhardt (commented):
Thanks! The code looks good, but unfortunately this is a breaking change and won't make it until Relm4 0.10 I'm afraid. Are you fine with using Relm4 from git or should we publish an early alpha release of Relm4 0.10 to make this work?

tymmesyde:
> Thanks! The code looks good, but unfortunately this is a breaking change and won't make it until Relm4 0.10 I'm afraid. Are you fine with using Relm4 from git or should we publish an early alpha release of Relm4 0.10 to make this work?

No problem, it can wait!

AaronErhardt (approved):

PR #663, opened 2024-07-24
Ignore sending error if async component was dropped quickly

nazar-pc:
#### Summary

As discussed on Matrix, this may cause panic if factory element is dropped too quickly

#### Checklist

- [ ] cargo fmt
- [ ] cargo clippy
- [ ] cargo test
- [ ] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,9 @@
 ## Unreleased
 
 ### Fixed
+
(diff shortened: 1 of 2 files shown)


AaronErhardt:
Looks good, but it would be great if you could add an entry to the changelog :)

nazar-pc:
Rebased on `main` and updated changelog

AaronErhardt (approved):
Thanks!

PR #682, opened 2024-09-13
Add `notify_filter_changed` function to typed views

ilya-zlobintsev:
#### Summary

This PR adds a `notify_filter_changed` function to all of the typed list abstractions. This is useful for making dynamic filters where the result may change between evaluations.

For example, I'm trying to make a text filter for a `TypedListView` where the user can filter items by entering text in a `SearchEntry`.

The filter looks like this:
```rust
        model.list_view.add_filter({
            let search_entry = widgets.search_entry.clone();
            move |item| {
                item
                    .name
                    .to_lowercase()
                    .contains(&search_entry.text().to_lowercase())
            }
        });
```

Currently to force this filter to update, you either have set it to inactive and then active, or completely re-add it to the list view on every input change.

With this function, I can simply notify that the filter has changed:
```rust
            AppMsg::FilterChanged(filter) => {
                if filter.is_empty() {
                    self.list_view.set_filter_status(0, false);
                } else {
                    self.list_view.set_filter_status(0, true);
                    self.list_view.notify_filter_changed(0);
                }
            }
```

#### Checklist

- [X] cargo fmt
- [X] cargo clippy
- [X] cargo test
- [X] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ### Changed
 
 + core: Add more async factory types to prelude
++ core: Add `notify_filter_changed` to typed views to allow dynamic filters
 + components: Hide recent button from `OpenButton` component when there is no entries
 
 ### Fixed
--- a/relm4/src/typed_view/column.rs
+++ b/relm4/src/typed_view/column.rs
@@ -309,9 +309,14 @@ where
 
     /// Notify that a certain filter has changed.
     /// This causes the filter expression to be re-evaluated.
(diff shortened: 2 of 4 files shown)


AaronErhardt (changes requested):
Thanks! Code looks good, but it would be nice to have a `bool` as returned type to indicate whether something happened, similar to `set_filter_status`. (Plus adding some docs about the return type would be great)

AaronErhardt (approved):
Thanks!

PR #692, opened 2024-12-22
Display "identical" background if initial icons are the same.

bobhy:
#### Summary

The "tracker" example does not display the green "identical" background if the 2 randomly-chosen icons happen to be identical when the app is first initialized.    

Repro:
1. `$ cargo run --example tracker`
2. Visually compare the displayed icons.  If they are not identical:
    a. exit the app
    b. go to step 1.

Expect (when the initially displayed icons are identical):
-- the app window background should be green, indicating identical

Observe:
-- the app window background is not green

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,10 @@
 
 + core: Add method `FactoryVecDeque::extend` to append multiple components efficiently.
 
+### Fixed
+
++ example "tracker": Display identical background when initial icons are in fact identical.
+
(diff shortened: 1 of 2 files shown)


bobhy:
I  don't think the clippy failures above have anything to do with my PR, but am willing to be educated...

AaronErhardt (approved):
Good catch, thanks!

PR #693, opened 2024-12-23
Accept clippy fixes

bobhy:
#### Summary
`cargo clippy` complaining about explicit lifetimes in code I haven't touched.
Happens both with current nightly (rustc 1.85.0-nightly (426d17342 2024-12-21))
and with current stable (rustc 1.83.0 (90b35a623 2024-11-26).

I'm also seeing some test failures as well, but only with stable.  So I don't expect this PR to pass CI.
#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test -- failures in stable, but not nightly
- [x] updated CHANGES.md -- n/a

### DIFF
--- a/relm4/src/factory/async/collections/vec_deque.rs
+++ b/relm4/src/factory/async/collections/vec_deque.rs
@@ -32,7 +32,7 @@ where
     inner: &'a mut AsyncFactoryVecDeque<C>,
 }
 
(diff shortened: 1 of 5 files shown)


AaronErhardt (approved):
Thanks!

PR #695, opened 2025-01-07
Reference Exercise Timer in README

mfep:
#### Summary

A trivial change to the README to add a project that employs Relm4. Kudos to all contributors of this library!


AaronErhardt (approved):
Thanks, your app looks really nice btw.!

I see that you have implemented translations and since this has been requested (and I don't have the expertise to add it myself) you could consider writing about your implementation in the Relm4 book (or maybe even the gtk-rs book) if you have time.

mfep:
Thank you, Aaron! Adding a section on i18n to the book would be a good idea. I just followed the method outlined in [relm4-template](https://github.com/Relm4/relm4-template), so by no means I can be considered an expert, but will see what I can do.

PR #698, opened 2025-01-22
add Words! to projects using Relm4

tronta:
#### Summary

I managed to create my first app with Relm4 and to publish it on Flathub.
There is still some things to do. But it's a lot of fun. Relm4 is really awesome!
Thank you everyone contributing to Relm4!!

#### Checklist

- [ ] cargo fmt
- [ ] cargo clippy
- [ ] cargo test
- [ ] updated CHANGES.md


AaronErhardt (approved):
Thanks! Looks like a really nice app, maybe I'll play a couple of rounds later today :)

tronta:
Thanks a lot. There is still a stupid bug where you cannot enter Z. I already pushed the 0.1.1 to Flathub which fixes it.
Unfortunately it's not yet rolled out. But I hope it will be in the next few hours.

PR #705, opened 2025-03-09
Add LACT to list of projects using Relm4 in README

ilya-zlobintsev:
#### Summary

Hi, my project has been using Relm4 for quite a while now, gradually moving over from plain gtk-rs. By now almost all of the UI uses Relm4, so I think it is fair to include it in the list.

#### Checklist
Not applicable


AaronErhardt (approved):
Thanks! I'm glad you enjoy using Relm4 :)

PR #706, opened 2025-03-09
Add Cargo.lock for CI usage, remove it from .gitignore

ilya-zlobintsev:
#### Summary

I've noticed that CI was failing due to one of the libraries getting an update that increases the minimum Rust version to one higher than what the CI uses (1.75).

While originally Rust discouraged committing Cargo.lock to library repositories, this recommendation was changed a while ago, in particular to avoid cases like this one regarding Rust versions:
https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

Committing Cargo.lock will make sure that library versions used in CI build won't get updated automatically, preventing the issue.

#### Checklist

This doesn't change anything for library end users, so I don't think a changelog entry is necessary 

- [ ] cargo fmt
- [ ] cargo clippy
- [ ] cargo test
- [ ] updated CHANGES.md


AaronErhardt:
Thanks a lot! I think the change makes sense, although it will inevitably bloat some git diffs. I guess that's still better than randomly failing CI jobs.

The one failing CI job is caused by a newer rustc version emitting different error messages (which are compared as part of the ui tests). We can simply ignore this for now or if you want, you can run the tests with `TRYBUILD=overwrite` to update the error output to the latest stable release.

ilya-zlobintsev:
I've updated the error text, CI should pass now.

AaronErhardt (approved):
Great, thanks a lot!

PR #713, opened 2025-03-10
Implement container traits for libpanel widgets

paulolieuthier:
#### Summary

Allow adding children to libpanel's widgets.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 + components: Add `icon` to `IconButton`
++ core: Implement container traits to allow easier usage of libpanel widgets in the view macro
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 + components: Add `icon` to `IconButton`
++ core: Support adding children to libpanel's widgets
```
AaronErhardt, line 5: This is a bit misleading. You could already add children to those widgets before, but now the extension traits are in place to make it more convenient. I would rather write something like "implement container traits to allow easier usage of libpanel widgets in the view macro".

AaronErhardt (changes requested):
Thanks! The code looks good, only the changelog could be more accurate.

AaronErhardt (approved):
Thanks a lot!

PR #740, opened 2025-04-24
Added an example of how to use a Stack with Relm4

0xfalafel:
#### Summary

Add an example `stack.rs` in the example directory.

This showcase:

* How to use a `Headerbar`.
* How to use a `Stack`.
* How to use a `StackSwitcher`.
* How to change the Stack displayed in the `update()` function.

I am still discovering Relm4, so I would love to see if there is a better way to do this.

#### Checklist

- [ ] cargo fmt
- [ ] cargo clippy
- [ ] cargo test
- [ ] updated CHANGES.md

### DIFF
--- a/examples/stack.rs
+++ b/examples/stack.rs
@@ -1,0 +1,117 @@
+use gtk::prelude::*;
+use relm4::{
+    component, gtk, ComponentParts, ComponentSender, RelmWidgetExt, SimpleComponent,
+};
+
+#[derive(Default)]
+struct StackApp {
+    stack: gtk::Stack,
+}
+
+#[derive(Debug)]
+enum StackAppMsg {
+    SwitchPage(u8),
+}
+
+#[component]
+impl SimpleComponent for StackApp {
+    type Init = ();
+    type Input = StackAppMsg;
+    type Output = ();
+    type Widgets = StackAppWidgets;
+
+    view! {
+        #[root]
+        gtk::ApplicationWindow {
+            set_default_size: (300, 200),
+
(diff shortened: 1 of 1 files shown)


examples/stack.rs
```diff
@@ -0,0 +1,117 @@
+use gtk::prelude::*;
+use relm4::{
+    component, gtk, ComponentParts, ComponentSender, RelmWidgetExt, SimpleComponent,
+};
+
+#[derive(Default)]
+struct StackApp {
+    stack: gtk::Stack,
```
AaronErhardt, line 8: Generally, you don't want to put widgets in your application state.

examples/stack.rs
```diff
@@ -0,0 +1,117 @@
...
+        model.stack = widgets.stack.clone();
+
+        ComponentParts { model, widgets }
+    }
+
+    fn update(&mut self, msg: Self::Input, _sender: ComponentSender<Self>) {
+        match msg {
+            StackAppMsg::SwitchPage(page) => {
+                println!("Go to page {}", page);
+                self.stack.set_visible_child_name(&page.to_string());
```
AaronErhardt, line 108: It's better to have a `visible_name: String` in your model and then use `'[watch]` in the view macro to update the child name. And if you really want to access the widget in the `update` function, there is [`update_with_view`](https://docs.rs/relm4/latest/relm4/component/trait.Component.html#method.update_with_view)

AaronErhardt (approved):
Thanks for the PR! There are a couple of things to improve, but the code is still pretty good overall.

PR #741, opened 2025-05-03
tracker.rs: Avoid repeats icon. 

M23SNEZ3:
Implement guaranteed icon uniqueness on refresh by adding regeneration validation.

#### Summary
Before this change, clicking the icon refresh button sometimes showed no visual change because the random generator could select the same icon consecutively.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [ ] updated CHANGES.md


AaronErhardt (approved):
Thanks, looks good.

PR #744, opened 2025-05-10
examples: Add AdwOverlaySplitView usage example

M23SNEZ3:
#### Summary
Add Libadwaita split layout example demonstrating AdwOverlaySplitView

Basic AdwOverlaySplitView implementation with:
   - Collapsible sidebar (toggleable via button)
   - Resizable sidebar width (toggles between 50px and 200px)
#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [ ] updated CHANGES.md

### DIFF
--- a/examples/split_layout.rs
+++ b/examples/split_layout.rs
@@ -1,7 +1,6 @@
-use relm4::prelude::*;
 use gtk::prelude::*;
 use relm4::adw::prelude::*;
-use tracker;
+use relm4::prelude::*;
 
 #[derive(Debug)]
 enum AppMsg {
@@ -12,7 +11,7 @@ enum AppMsg {
 #[tracker::track]
 struct AppModel {
     sidebar_width: i32,
-    sidebar_visible:  bool,
+    sidebar_visible: bool,
 }
 
 #[relm4::component]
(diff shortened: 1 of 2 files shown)


examples/split_layout.rs
```diff
@@ -0,0 +1,109 @@
...
+    type Input = AppMsg;
+    type Output = ();
+
+    view! {
+        adw::ApplicationWindow {
+            set_default_width: 600,
+            set_default_height: 400,
+
+            #[wrap(Some)]
+            set_content = &adw::OverlaySplitView
```
AaronErhardt, line 29:
```suggestion
            adw::OverlaySplitView
```

AaronErhardt (approved):
Code looks good. Technically, you could use the shorter syntax for adding new widgets more often, but that's to some degree a personal preference. So if you want, we can merge this as is.

M23SNEZ3:
Okay, I'll be in the loop in a future. But I think my code already clear and I don't want to change something. Therefore, if it's possible, I would like to merge without changes:)

AaronErhardt:
Thanks!

PR #797, opened 2025-11-01
feat: add `AsyncReducer` with `AsyncReducible`

muni-corn:
#### Summary

Adds `async` variants of `Reducer` and `Reducible`.

Use case: the Bluetooth service for my custom desktop shell, where a global state is most efficient, and it can be reduced via events, but `async`/`await` is needed: https://github.com/muni-corn/cadenza-shell/blob/077197513aef270f025cfb955dd93f220834aa67/src/bluetooth.rs

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
++ core: Add `AsyncReducible` and `AsyncReducer` for reducers that can use `async`/`await`
+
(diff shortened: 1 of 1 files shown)


AaronErhardt (approved):
Thanks! Code looks solid.

PR #820, opened 2025-12-16
message_from_grid_view.rs: Fix incorrect output when there are a large number of objects

M23SNEZ3:
### Summary
When objects more than 200-300, output does not work correctly. This PR solved this problem.

### For example
When you click on the 150 number, click does not trigger correctly, and output looks like:
``` 
Name: "mmstick"
Name: "tronta"
Name: "mmstick"
Name: "tronta"
```
This problem is solved with `glib::SignalHandlerId`. I have no idea why it works, but... It just works....
### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [ ] updated CHANGES.md

### DIFF
--- a/examples/message_from_grid_view.rs
+++ b/examples/message_from_grid_view.rs
@@ -1,6 +1,7 @@
+use glib::SignalHandlerId;
 use gtk::prelude::*;
 use rand::seq::IteratorRandom;
-use relm4::gtk::{glib, glib::SignalHandlerId};
+use relm4::gtk::glib;
 use relm4::{
     prelude::*,
     typed_view::grid::{RelmGridItem, TypedGridView},
@@ -22,10 +23,11 @@ fn random_name() -> &'static str {
         .expect("Could not choose a random name")
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 struct MyGridItem {
     name: &'static str,
     sender: ComponentSender<App>,
+    button_click_handler_id: Option<SignalHandlerId>,
 }
 
 impl MyGridItem {
@@ -33,6 +35,7 @@ impl MyGridItem {
         Self {
             name: random_name(),
             sender,
+            button_click_handler_id: Default::default(),
         }
     }
 }
@@ -40,7 +43,6 @@ impl MyGridItem {
 struct Widgets {
     button: gtk::Button,
     label: gtk::Label,
-    click_handler: Option<SignalHandlerId>,
 }
 
 impl RelmGridItem for MyGridItem {
@@ -65,22 +67,15 @@ impl RelmGridItem for MyGridItem {
             }
         }
 
-        let widgets = Widgets {
-            label,
-            button,
-            click_handler: None,
-        };
+        let widgets = Widgets { label, button };
 
         (my_box, widgets)
     }
 
     fn bind(&mut self, widgets: &mut Self::Widgets, _root: &mut Self::Root) {
         widgets.label.set_label(self.name);
-        if let Some(id) = widgets.click_handler.take() {
-            widgets.button.disconnect(id);
-        }
         let name = self.name;
-        let handler_id = widgets.button.connect_clicked(glib::clone!(
+        let button_click_handler_id = widgets.button.connect_clicked(glib::clone!(
             #[strong(rename_to = sender)]
             self.sender,
             move |_btn| {
@@ -88,7 +83,14 @@ impl RelmGridItem for MyGridItem {
                 sender.input(Msg::Print(name));
             }
         ));
-        widgets.click_handler = Some(handler_id);
+        self.button_click_handler_id
+            .replace(button_click_handler_id);
+    }
+
+    fn unbind(&mut self, widgets: &mut Self::Widgets, _root: &mut Self::Root) {
+        if let Some(id) = self.button_click_handler_id.take() {
+            widgets.button.disconnect(id)
+        }
     }
 }
 


zekefast:
@M23SNEZ3 This issue is kinda interesting, but also somewhat very typical. So, it was my fault that I didn't spot it during review of #806 as I was mostly focused on cleaning up the `.clone()` calls and making sure we don't leak the memory by cycle-referencing widgets.

As [`TypedGridView`](https://github.com/Relm4/Relm4/blob/main/relm4/src/typed_view/grid.rs#L180) under the hood uses [`gtk4::GridView`](https://gtk-rs.org/gtk4-rs/git/docs/gtk4/struct.GridView.html) it binds and unbinds data to list items as you scroll the [`gtk4::ScrolledWindow`](https://gtk-rs.org/gtk4-rs/git/docs/gtk4/struct.ScrolledWindow.html).

You can make sure in that just by putting some `println!("Binded!")` into example's `MyGridItem::bind` method, e.g. just before the call to `.connect_clicked()` on button. Don't put it inside the click handler as you don't see the binding effect than!
The tricky part is when binding/unbinding occurs. For small sets of items inside [`gtk4::GridView`](https://gtk-rs.org/gtk4-rs/git/docs/gtk4/struct.GridView.html) they all binded immediately and nothing is unbinded. But I notice once I add 100 of buttons, binded only 97, and 3 is waiting to be binded.

So, here is why the unexpected effect of multi-line out happens and why click handler is called several times. We have no unbind handler. We inherit [empty implementation from `RelmGridItem` trait](https://github.com/Relm4/Relm4/blob/main/relm4/src/typed_view/grid.rs#L28). When we add many items to the grid then binding/unbinding process starts. As we attach click handler inside `bind` function

```rust
    fn bind(&mut self, widgets: &mut Self::Widgets, _root: &mut Self::Root) {
        widgets.label.set_label(self.name);
        let name = self.name;
        widgets.button.connect_clicked(glib::clone!(
            #[strong(rename_to = sender)]
            self.sender,
            move |_btn| {
                // Use the cloned sender to send a message
                sender.input(Msg::Print(name));
            }
        ));
    }
```
on each `bind` call we attach additional click handler to the button, but we've never remove them. Hence, unexpected behavior of multiple calls of click handler, because when actual click occurs it triggers several handlers.

The solution is to attach handler only once and than somehow pass data into it or connect/disconnect handler as @M23SNEZ3 have been showed in this PR already. Or just avoid using `GridView` (i.e. `TypedGridView`) if amount of data you work with isn't large enough (assume you still want to showcase work with the grid here).

I think it make more sense to store [`SingnalHandlerId`](https://gtk-rs.org/gtk-rs-core/stable/0.21/docs/glib/signal/struct.SignalHandlerId.html) inside `MyGridItem` instead of `Widgets`. I prepared some changes to the original example to showcase how I would do it if we choose to go [`SingnalHandlerId`](https://gtk-rs.org/gtk-rs-core/stable/0.21/docs/glib/signal/struct.SignalHandlerId.html) route.

```diff
diff --git a/examples/message_from_grid_view.rs b/examples/message_from_grid_view.rs
index c0e087a..bcb30a1 100644
--- a/examples/message_from_grid_view.rs
+++ b/examples/message_from_grid_view.rs
@@ -1,3 +1,4 @@
+use glib::SignalHandlerId;
 use gtk::prelude::*;
 use rand::seq::IteratorRandom;
 use relm4::gtk::glib;
@@ -22,10 +23,11 @@ fn random_name() -> &'static str {
         .expect("Could not choose a random name")
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 struct MyGridItem {
     name: &'static str,
     sender: ComponentSender<App>,
+    button_click_handler_id: Option<SignalHandlerId>,
 }
 
 impl MyGridItem {
@@ -33,6 +35,7 @@ impl MyGridItem {
         Self {
             name: random_name(),
             sender,
+            button_click_handler_id: Default::default(),
         }
     }
 }
@@ -72,7 +75,7 @@ impl RelmGridItem for MyGridItem {
     fn bind(&mut self, widgets: &mut Self::Widgets, _root: &mut Self::Root) {
         widgets.label.set_label(self.name);
         let name = self.name;
-        widgets.button.connect_clicked(glib::clone!(
+        let button_click_handler_id = widgets.button.connect_clicked(glib::clone!(
             #[strong(rename_to = sender)]
             self.sender,
             move |_btn| {
@@ -80,6 +83,14 @@ impl RelmGridItem for MyGridItem {
                 sender.input(Msg::Print(name));
             }
         ));
+        self.button_click_handler_id
+            .replace(button_click_handler_id);
+    }
+
+    fn unbind(&mut self, widgets: &mut Self::Widgets, _root: &mut Self::Root) {
+        if let Some(id) = self.button_click_handler_id.take() {
+            widgets.button.disconnect(id)
+        }
     }
 }

```

As usually you can apply it with `git apply <path_to_the_diff_file>` or with `patch -p1 < <path_to_the_diff_file>`.

Notice, `MyGridItem` can't be cloned as [`SingnalHandlerId`](https://gtk-rs.org/gtk-rs-core/stable/0.21/docs/glib/signal/struct.SignalHandlerId.html) does not implement [`Clone`](https://doc.rust-lang.org/std/clone/trait.Clone.html) trait, so you can't derive [`Clone`](https://doc.rust-lang.org/std/clone/trait.Clone.html) for `MyGridItem`. From the docs:

> This type does not implement Clone to prevent disconnecting the same signal handler multiple times.

For the first idea with connecting click handler only once per button I could imagine the following solution. 
1. Assign click handler to button in `setup` hook
2. Pass weak reference to `label` inside click handler
3. Retrieve text from `label` on the button in the handler and send it in the `Msg::Print()`. You'll have to copy as both [`Label.label`](https://docs.rs/gtk4/latest/i686-pc-windows-msvc/gtk4/struct.Label.html#method.label) and [`Label.text`](https://docs.rs/gtk4/latest/i686-pc-windows-msvc/gtk4/struct.Label.html#method.text) returns owned value of [`GString`](https://docs.rs/glib/0.21.5/i686-pc-windows-msvc/glib/gstring/struct.GString.html).
4. In `bind` hook change the text on `Label` as usual. As click handler contains reference to `Label` when it will be triggered it gets new text.
5. Ideally we still have to do data unbind, i.e. remove text from labels as on large number of items lack of unbinding will eat the RAM.

All of that got me thinking that we might use GridView incorrectly as with large number of items instead of bind/unbinding data from the model we hold data on the MyGridItem which lead to slowdown and memory consumption. But I still have to think about it and better understand how `TypedGridView` works with Model.

@M23SNEZ3 I suggest you to read [List Widgets](https://gtk-rs.org/gtk4-rs/stable/latest/book/list_widgets.html) chapter if you haven't yet. They have a good description of bind/unbind for View widgets.
Also overview of [List Widgets](https://docs.gtk.org/gtk4/section-list-widget.html) might be handy as well.

PR #904, opened 2026-05-02
Document `update_view` requirement on factory methods

Lynnesbian:
#### Summary

Documents the need to call `update_view` in `update_with_view` and `update_cmd_with_view` on sync and async factories, bringing them in line with [the equivalent `Component` docs](https://docs.rs/relm4/0.10.0/relm4/component/trait.Component.html#method.update_with_view).

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md


AaronErhardt (approved):
Thanks!

PR #915, opened 2026-05-12
Allow setting conditional widget stack properties

### LINKED ISSUE #624: Allow customizing stack properties
Currently when using `match` the only thing that can be customized is transition type, but it would be great to set other properties like `vhomogeneous`.

It is possible to work around with `#[name = "name"]` in components, but in factories it is even less convenient than that.

Lynnesbian:
#### Summary

Allows setting conditional widget stack properties:

```rust
view! {
	if model.foo {
		gtk::Label { set_label: "foo" }
	} else {
		gtk::Label { set_label: "bar" }
	} -> {
		#[watch]
		set_halign: model.stack_align;

		set_hhomogeneous: false,
	}
}
```

Fixes #624.

This is my first contribution to the `view` macro, so I might not have done things optimally.

There's a lot to discuss before this is merged:

- Is this the right syntax? It's similar to the returned widget syntax, but doesn't allow specifying a name or type.
- Do we want to expose *all* of the `gtk::Stack`'s properties? Things like [`add_child`](https://docs.gtk.org/gtk4/method.Stack.add_child.html) and [`remove`](https://docs.gtk.org/gtk4/method.Stack.remove.html) could cause issues, *especially* the latter.
- Should we keep `#[transition]`, given that this PR makes it possible to set the `Stack`'s transition property directly?

If this is merged, the book will need to be updated.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 + docs: Document `update_view` requirement on factory methods
 + core: Implement container traits for `gtk::TreeExpander`, `gtk::MenuButton`, `gtk::SearchBar`, `gtk::Viewport`, and
   `adw::Dialog`
++ macros: Allow setting conditional widget `gtk::Stack` properties
 
 ### Changed
 
--- a/examples/README.md
+++ b/examples/README.md
@@ -39,6 +39,12 @@ keyboard shortcuts and [menus](https://docs.rs/gio/latest/gio/struct.Menu.html).
 
 A [`MessageDialog`](https://docs.rs/gtk4/latest/gtk4/struct.MessageDialog.html) made of multiple [components].
 
+### `conditional_properties.rs`
+
+Set properties on a
+[conditional widget](https://relm4.org/book/stable/component_macro/reference.html#conditional-widgets)'s
+[`Stack`](https://docs.rs/gtk4/latest/gtk4/struct.Stack.html).
+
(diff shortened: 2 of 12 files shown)


AaronErhardt (approved):
Looks good! Thanks for working on this, I think you might be the first person to dare going deeper into the macro code. Well done!

AaronErhardt:
Sorry, I just jumped at looking at the code and overlooked the questions.

> Is this the right syntax? It's similar to the returned widget syntax, but doesn't allow specifying a name or type.

I think there are few other options for exposing this kind of functionality. And it makes sense to re-use the returned widget syntax, because you could argue that the stack is returned by the macro from the if-else block.

> Do we want to expose all of the gtk::Stack's properties? Things like [add_child](https://docs.gtk.org/gtk4/method.Stack.add_child.html) and [remove](https://docs.gtk.org/gtk4/method.Stack.remove.html) could cause issues, especially the latter.

I don't think it would make sense to saveguard the macro. It is unlikely to be abused IMO.

> Should we keep #[transition], given that this PR makes it possible to set the Stack's transition property directly?

I think we should probably deprecate transition because it would no longer be necessary and it's odd to have a shorthand just for this property alone.

PR #916, opened 2026-05-13
Deprecate `#[transition]` on conditional widgets

Lynnesbian:
#### Summary

Deprecates the `#[transition]` conditional widget attribute as suggested in [this comment on #915](https://github.com/Relm4/Relm4/pull/915#issuecomment-4444190004).

If this gets merged, I'll update the book to show the new method.

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/relm4/src/lib.rs
+++ b/relm4/src/lib.rs
@@ -43,6 +43,8 @@ pub mod binding;
 pub mod component;
 pub mod factory;
 pub mod loading_widgets;
+#[doc(hidden)]
+pub mod macro_helper;
 pub mod shared_state;
 pub mod typed_view;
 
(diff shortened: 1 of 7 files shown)


relm4/src/lib.rs
```diff
@@ -252,3 +252,9 @@ pub fn set_global_css_from_file_with_priority<P: AsRef<std::path::Path>>(
 pub fn set_global_css_from_file<P: AsRef<std::path::Path>>(path: P) -> Result<(), std::io::Error> {
     set_global_css_from_file_with_priority(path, gtk::STYLE_PROVIDER_PRIORITY_APPLICATION)
 }
+
+/// Dummy function for marking the `#[transition]` macro attribute as deprecated.
+/// See https://stackoverflow.com/a/77267752.
+#[deprecated = "The `#[transition]` macro attribute is deprecated. Call `set_transition_type` on the returned widget instead."]
+#[doc(hidden)]
+pub const fn transition() {}
```
AaronErhardt: I don't like having this in the root of the crate with this quite generic name. The best would probably having a hidden `macro_helper` module.

AaronErhardt (approved):
Looks good, thanks!

PR #919, opened 2026-05-19
Conditional root widgets

### LINKED ISSUE #592: Support `match` at the top root of the widget
`match` is expanded to a `Stack`, which I think should be a perfectly valid root element, but `view!` macro doesn't like it right now in that context.

Workaround is to simply wrap it in another `Box`, but it feels unnecessary and adds another unnecessary layer of nesting.

### LINKED ISSUE #690: Assign conditional or factory builder output to root element in view! macro
When working with a factory builder in the `init` or `init_model`, the output must be assigned as a child to a parent widget, e.g. a `gtk::Box`. Consider this example of a `FactoryComponent` view:

```rust
view! {
    #[root]
    gtk::Box { // this widget is only needed because of constraints of the `view!` macro
        // set_spacing: 10,
        self.props.widget() ->  // cannot be assigned directly to #[root] 
        &gtk::Box {
            set_spacing: 10,
        },
    },
```
In a similar fashion, for a conditional widget I must wrap the output of the conditional `if` or `match` in an outer widget (`gtk::Box`) and assign the value using `append`:
```rust
view! {
    #[root]
    gtk::Box {  // this widget is only needed because of constraints of the `view!` macro
        set_orientation: gtk::Orientation::Horizontal,
        set_spacing: 10,
        append = match &self.item.value { // cannot be assigned directly to #[root]
            PropertyValue::Text(s) => {
                gtk::Box {
                    set_orientation: gtk::Orientation::Horizontal,
                    set_spacing: 10,
                    ...
```
In both of these cases we are creating superfluous boxes.

It would be great if we could assign the conditional output and the output of a builder in an `init` or `init-model` method directly to the root element as this would eliminate the need for the intermediate and superfluous widgets.

For more detail on the reason behind this suggestion, please refer to the support issue [Factory for creating rows in a table](https://github.com/orgs/Relm4/discussions/689)

Lynnesbian:
#### Summary

Allows [conditional widgets](https://relm4.org/book/stable/component_macro/reference.html#conditional-widgets) to be used as root widgets. Fixes #690, fixes #592.

I'm happy with the syntax and it seems to be functioning well, but I'd appreciate if someone with experience in how the macros work took a closer look at things to make sure I've got it right. I've tested it with both `if` and `match`, names (`name = if ...`), the `#[root]` annotation, and the conditional property syntax from #915.

#### Example
```rust
view! {
    if model.state {
        gtk::Label {
            #[watch]
            set_label: &format!("Toggled {} times", model.counter),
        }
    } else {
        gtk::Label {
            set_label: "False",
        }
    }
}
```

#### Checklist

- [x] cargo fmt
- [x] cargo clippy
- [x] cargo test
- [x] updated CHANGES.md

### DIFF
--- a/relm4-macros/src/widgets/codegen/conditional_init.rs
+++ b/relm4-macros/src/widgets/codegen/conditional_init.rs
@@ -76,6 +76,18 @@ impl Widget {
 }
 
 impl ConditionalWidget {
+    pub(crate) fn init_conditional_init_stream(
+        &self,
+        stream: &mut TokenStream2,
+        model_name: &Ident,
+    ) {
+        if let Some(properties) = &self.properties {
+            let w_name = &self.name;
+            properties.conditional_init_stream(stream, w_name, model_name, true);
+        }
+        self.conditional_init_stream(stream, model_name);
+    }
+
     fn conditional_init_stream(&self, stream: &mut TokenStream2, model_name: &Ident) {
         let brach_stream = match &self.branches {
             ConditionalBranches::If(if_branches) => {
--- a/relm4-macros/src/widgets/codegen/util/widget.rs
+++ b/relm4-macros/src/widgets/codegen/util/widget.rs
@@ -1,4 +1,4 @@
-use crate::widgets::ViewWidgets;
+use crate::widgets::{TopLevelInner, ViewWidgets};
 
 impl ViewWidgets {
     /// Get a mutable reference to the root widget
@@ -7,8 +7,9 @@ impl ViewWidgets {
             .top_level_widgets
             .iter_mut()
             .find(|w| w.root_attr.is_some())
+            && let TopLevelInner::Widget(widget) = &mut root_widget.inner
         {
-            root_widget.inner.name_assigned_by_user = true;
+            widget.name_assigned_by_user = true;
         }
     }
 }
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
 + core: Implement container traits for `gtk::TreeExpander`, `gtk::MenuButton`, `gtk::SearchBar`, `gtk::Viewport`, and
   `adw::Dialog`
 + macros: Allow setting conditional widget `gtk::Stack` properties
++ macros: Conditional root widgets
 
 ### Changed
 
--- a/examples/README.md
+++ b/examples/README.md
@@ -45,6 +45,11 @@ Set properties on a
 [conditional widget](https://relm4.org/book/stable/component_macro/reference.html#conditional-widgets)'s
 [`Stack`](https://docs.rs/gtk4/latest/gtk4/struct.Stack.html).
 
+### `conditional_root.rs`
+
+Use a [conditional widget](https://relm4.org/book/stable/component_macro/reference.html#conditional-widgets) as a
+[component][components]'s root widget.
+
 ### `data_binding.rs`
 
 Automatically update the UI using [data binding](https://docs.rs/relm4/latest/relm4/binding/index.html).
--- a/examples/conditional_root.rs
+++ b/examples/conditional_root.rs
@@ -1,0 +1,102 @@
+use relm4::gtk::prelude::*;
+use relm4::prelude::*;
+pub struct App(Controller<Inner>);
+
+#[relm4::component(pub)]
+impl SimpleComponent for App {
+    type Init = ();
+    type Input = ();
+    type Output = ();
+
+    view! {
+        #[root]
+        window = gtk::Window {
+            set_title: Some("Conditional Root Example"),
+            set_default_size: (200, 150),
+
+            gtk::Box {
+                set_orientation: gtk::Orientation::Vertical,
+                set_spacing: 5,
+                set_margin_all: 5,
+                set_align: gtk::Align::Center,
+
+                #[local_ref]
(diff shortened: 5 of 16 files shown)


relm4-macros/src/widgets/codegen/util/widget.rs
```diff
@@ -7,8 +7,9 @@ impl ViewWidgets {
             .top_level_widgets
             .iter_mut()
             .find(|w| w.root_attr.is_some())
+            && let TopLevelInner::Widget(widget) = &mut root_widget.inner
```
AaronErhardt, line 10: Why would this not apply to ConditionalWidget?

relm4-macros/src/widgets/codegen/conditional_init.rs
```diff
@@ -76,7 +76,21 @@ impl Widget {
...
+        if let Some(properties) = &self.properties {
+            let w_name = &self.name;
+            properties.conditional_init_stream(stream, w_name, model_name, true);
+        }
+        self.conditional_init_stream(stream, model_name);
+    }
+
     fn conditional_init_stream(&self, stream: &mut TokenStream2, model_name: &Ident) {
+        stream.extend(quote! {#[doc = "conditional_init_stream"]});
+
```
AaronErhardt: Why the doc attribute here?

AaronErhardt (commented):
Looks good overall. It adds a bit more complexity than I'd like, but all seems reasonable and mostly contained.

relm4-macros/src/widgets/codegen/conditional_init.rs
Lynnesbian: Ah, my bad, I added some doc comments while I was trying to understand how the macro worked and forgot to remove this one

relm4-macros/src/widgets/codegen/util/widget.rs
Lynnesbian: I'm not entirely sure what this does. I can see it's setting the `name_assigned_by_user` property, which is only read by [the `has_struct_field` method on `Widget`](https://github.com/Lynnesbian/Relm4/blob/2f31359b55cbbc332f30ef74cd3ef7ec2078a578/relm4-macros/src/widgets/codegen/util/has_struct_field.rs#L6-L14). `ConditionalWidget` has no equivalent method.

If I did update the code to set this on `ConditionalWidget`, would I need to iterate through the `ConditionalBranches` and set the property on each one, or would I need to add a `name_assigned_by_user` property to the `ConditionalWidget` itself? I'm not really sure how to go about this.

AaronErhardt: For constancy, now that conditional widgets can have properties, I guess it would make sense to add the `name_assigned_by_user` field there, too. The idea was, if I remember correctly, to avoid dead code and unnecessary pollution of the scope by avoiding bringing variables into scope that are not needed. Any widget that does not update it's properties isn't needed, unless the user assigned a name because then it might be used somewhere else in the code. The other option would be drop this entirely, because it is not a big deal to generate a tiny bit of dead code, especially if it reduces the complexity of the macro code.

For reference, this was the commit that added this: https://github.com/Lynnesbian/Relm4/commit/8d92731c03e01cb6a260f53e683ced38db3210d2

Lynnesbian: Thanks for the explanation and especially the commit adding the function, that helped a lot with tracking it down.

Implementing the dead code logic with `name_assigned_by_user` breaks the conditional widget's update logic. `update_view_stream` always emits code for controlling the `Stack`'s current page ([as can be seen here](https://github.com/Lynnesbian/Relm4/blob/673bf3336f8d674dc8343c222bdcb516f1a78356/relm4-macros/src/widgets/codegen/update_view.rs#L159-L163)). 

`Widget` calls `has_struct_field` in three locations, but adding a similar check to `ConditionalWidget` in any one of them breaks the page-change code:

- [`return_stream`](https://github.com/Lynnesbian/Relm4/blob/673bf3336f8d674dc8343c222bdcb516f1a78356/relm4-macros/src/widgets/codegen/return_fields.rs#L49): The `Stack` is no longer returned from the `init` function
- [`destructure_stream`](https://github.com/Lynnesbian/Relm4/blob/673bf3336f8d674dc8343c222bdcb516f1a78356/relm4-macros/src/widgets/codegen/destructure_fields.rs#L49): The `Stack` is stored in the `Self::Widgets` struct, but `update_with_view` doesn't name it
- [`struct_fields_stream`](https://github.com/Lynnesbian/Relm4/blob/673bf3336f8d674dc8343c222bdcb516f1a78356/relm4-macros/src/widgets/codegen/struct_fields.rs#L60): The `Stack` is never stored in `Self::Widgets`

Any one of these three renders `update_with_view` unable to change the `Stack`'s page (and causes compile errors).

I believe the only way this can work is if the component is initialised but never updated, which mostly defeats the purpose of using a conditional widget.

AaronErhardt: Thanks for having a closer look. I think it should be good as is and if needed, we can still investigate the potential for optimizations.

AaronErhardt (approved):
Looks good overall. It adds a bit more complexity than I'd like, but all seems reasonable and mostly contained.

